PR #2, opened 2020-05-19
Hello world 🖖

Yajo:
PR from same repo, to develop CI.

This closes #1.


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,33 @@
+name: automerge
```
Yajo, line 1: @sbidoul This workflow definition lets us just add an `automerge` label to the PR and let it be automatically merged when all required CI checks pass.

I don't know if oca git bot is applicable for this repo, given is different nature. That's why I added this, plus it's something I almost do everywhere these days.

If it is redundant with our git bot and you think it would confuse or would be not of any help, I can remove this. You're the pro there, just tell me please.

sbidoul: > If it is redundant with our git bot and you think it would confuse or would be not of any help

A bit of all this, yes ;) I think ocabot merge nobump will work just fine here too.

sbidoul:
> mark the pre-commit and test workflows as required to merge

Done

README.md
```diff
@@ -0,0 +1,75 @@
...
+You might find some things that you can reuse in your own templates, but in general
+terms this template is not meant to support being used as is for other organizations.
+
+## The legal stuff
+
+Copyright holder: [Odoo Community Association](https://odoo-community.org/).
+
+Template license: [MIT](LICENSE)
+
+License of the rendered repositories: [AGPL](LICENSE.jinja)
```
sbidoul, line 75: The repos themselve don't really have a license, since each addon can have a different one.

.pre-commit-config.yaml.jinja
```diff
@@ -0,0 +1,97 @@
...
+  # Files and folders generated by bots, to avoid loops
+  ^setup/|/static/description/index\.html$|
+  # Maybe reactivate this when all README files include prettier ignore tags?
+  ^README\.md$|
+  # Library files can have extraneous formatting (even minimized)
+  /static/(src/)?lib/|
+  # Repos using Sphinx to generate docs don't need prettying
+  ^docs/_templates/.*\.html$|
+  # You don't usually want a bot to modify your legal texts
+  (LICENSE.*|COPYING.*)
```
sbidoul, line 12: One thing we'll want is a mechanism to exclude uninstallable addons. @ThomasBinsfeld has a pre-commit hook that does that. This can be done in a follow-up.

README.md
Yajo: But they do have... example: https://github.com/OCA/server-tools/blob/12.0/LICENSE

.pre-commit-config.yaml.jinja
Yajo: OK, although AFAIK nowadays all new branches start empty, so there should be no uninstallable addons there.

.travis.yml.jinja
```diff
@@ -0,0 +1,82 @@
...
+    - stage: test
+      env:
+        - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
+    - stage: test
+      env:
+        - TESTS="1" ODOO_REPO="OCA/OCB"
+    {%- endif %}
+env:
+  global:
+    - VERSION="{{ odoo_version }}" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
```
sbidoul, line 67: MQT now has an option to [install dependencies with pip](https://github.com/OCA/maintainer-quality-tools/blob/67d3880b085a99d820e5a65cc7354dae95d1b4b0/travis/travis_install_nightly#L119-L134) (env var `MQT_DEP="PIP"`). We are experimenting that in some repos. Can it be part of the template?

README.md
```diff
@@ -0,0 +1,75 @@
...
+# Update the repo
+cd some-repo
+copier update
+# Reformat updated files
+pre-commit run
+# Commit update
+git commit -am 'Updated from template'
+# Reformat all other files, in case some pre-commit configuration was updated
+pre-commit run -a || git commit -am 'Reformatted after template update'
+```
```
sbidoul, line 50: To have the possibility to apply this massively to all repos on the 13.0 branch, I think we need a mechanism to opt-out the generation of `.travis.yml` and `README.md`. For the other files it should be ok.

.travis.yml.jinja
Yajo: Yes, of course. Let me check that.

README.md
Yajo: Yes, you can add `--skip` to your copier invocation. That's supported.

README.md
sbidoul: Yeah, but that's confusing. Look for instance in the `queue_job` repo which has a mix of LGPL and AGPL addons, but the repo file says AGPL.

.gitignore
```diff
@@ -0,0 +1,79 @@
...
+htmlcov/
+.tox/
+.coverage
+.cache
+nosetests.xml
+coverage.xml
+
+# Translations
+*.mo
+*.pot
```
sbidoul, line 45: We commit .pot files in git.

.gitignore
```diff
@@ -0,0 +1,79 @@
+# Project specific
+oca.cfg
```
sbidoul, line 2: What is this oca.cfg?

.isort.cfg
```diff
@@ -0,0 +1,12 @@
...
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+combine_as_imports=True
+use_parentheses=True
+line_length=88
+known_odoo=odoo
+known_odoo_addons=odoo.addons
+sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
+known_third_party=copier,plumbum,pytest,yaml
```
sbidoul, line 12: We don't want to update repos `.isort.cfg` with this line.

.pre-commit-config.yaml.jinja
```diff
@@ -0,0 +1,97 @@
...
+  # TODO Avoid awebdeveloper/pre-commit-prettier if possible
+  # HACK https://github.com/prettier/prettier/issues/7407
+  - repo: https://github.com/awebdeveloper/pre-commit-prettier
+    rev: v0.0.1
+    hooks:
+      - id: prettier
+        name: prettier xml plugin
+        additional_dependencies:
+          - "prettier@1.19.1"
+          - "@prettier/plugin-xml@0.5.0"
```
sbidoul, line 33:
```suggestion
          - "@prettier/plugin-xml@0.7.2"
```

.prettierrc.yml
```diff
@@ -0,0 +1,7 @@
+# Defaults for all prettier-supported languages.
+# Prettier will complete this with settings from .editorconfig file.
+bracketSpacing: false
+printWidth: 88
+proseWrap: always
+semi: true
+trailingComma: "es5"
```
sbidoul, line 7:
```suggestion
trailingComma: "es5"
xmlWhitespaceSensitivity: "ignore"
```

.travis.yml.jinja
```diff
@@ -0,0 +1,82 @@
+language: python
+cache:
+  directories:
+    - $HOME/.cache
```
sbidoul, line 4:
```suggestion
    - $HOME/.cache
    - $HOME/.cache/pre-commit
```

.gitignore
Yajo: No idea, but it was added in https://github.com/OCA/maintainer-quality-tools/pull/518. Should I remove it?

cc @MiquelRForgeFlow

.isort.cfg
Yajo: Ooops...

setup.cfg
```diff
@@ -0,0 +1,3 @@
+[tool:pytest]
+addopts = -nauto
+norecursedirs = vendor
```
sbidoul, line 3: Need to exclude this from copier files, or move it to pyproject.toml if pytest supports that.

.travis.yml.jinja
Yajo: Ain't it better to cache all cache?

sbidoul: Ah yes, caching all the cache directory. I don't know if that could be too aggressive. We usually have .cache/pip and .cache/pre-commit.

README.md
sbidoul: That'd be `--exclude` ? I tested that with `--exclude README.md --exclude .travis.yml` but the it started overwriting my `.git` directory.

.isort.cfg
Yajo: Oh my local MQT clone was out of sync, you'll see some failures due to that. Wait for next push please, sorry.

copier.yml
```diff
@@ -0,0 +1,83 @@
...
+
+repo_name:
+  type: str
+  help:
+    Tell me the project name. It's supposed to be human-readable. So, server-tools
+    project could be named like "Tools for server environment(s)"
+
+repo_description:
+  type: str
+  help: Please write a short description about what this repo is about.
```
sbidoul, line 46: Can we have a rich markdown description?

README.md
Yajo: Yes, I don't quite understand that... Might be the licensing for files outside of a module? We can take this chance to state clearly in the README that each module has its own license. :shrug:

README.md
Yajo: `--skip` skips the file only if it is present, while `--exclude` totally ignores it. `--skip` would apply in this case.

Also when there is any kind of diff, if ran interactively, copier asks you if you want to skip a file. However `--skip` is still useful to script it.

.gitignore
Yajo: I'll remove this one, as I see no use for it. Ping me otherwise please.

README.md
sbidoul: If I read well `--skip` applies to all files? So how to update, say `.pre-commit-config.yaml` while not updating `.travis.yml`?

.travis.yml.jinja
Yajo: OK I was wondering why wasn't done like this, but nevermind, you're right, I'll better leave it as it was.

copier.yml
Yajo: Right now multiline input is not supported in copier, at least not interactively: https://github.com/pykong/copier/issues/210

However, you can pass multiline input in `--data`.

Also, you can write any text you want as long as it's one line, which includes any markdown markup.

README.md
Yajo: I'm thinking... Is it actually legal to combine modules licensed with, say, Apache-2.0, in a repo that is itself licensed under AGPL-3?

Maybe we should license our repositories under MIT or something like that? :thinking: 

Are we allowed to do such change at this point of history? Or better for v14+?

Yajo: These days I'm using https://choosealicense.com/licenses/bsl-1.0/ which is "Apache but shorter", could be a good option too.

Yajo:
All should be resolved now, except the points that are unresolved. :partying_face: http://instantrimshot.com/

README.md
sbidoul: I think the current situation is just because most people don't pay attention to that.
I'd say it's totally allowed to mix licences in one repo, as long as the license for each file is clear.
I'm not sure it's important to have a specific license for these dotfiles at the repo root.

sbidoul:
Looks promising. Thanks for this work!

I unresolved the skip/exclude thing, though. It's unclear to me how that would work.

README.md
Yajo: Well, never had to do that, but if I'm not wrong, this should work:

```bash
copier --skip "**" --skip "!.pre-commit-config.yaml" -f
```

README.md
Yajo: Well, AFAIK you're not allowed to mix any \*GPL license with code that doesn't respect the fundamental rights they enforce. At least not the other thing depending on the \*GPL thing, which would be our case. Except for LGPL.

For <= Odoo v8 this wouldn't matter because the base was AGPL, so anything built on it had to be AGPL too.

For Odoo v9+ it can be a problem. But changing this at this point will lead to an endless discussion with all the community.

I think we should just start using another license for v14+ repos. Let's leave things as they are now.

Yajo:
Yes, exclude is: Copier, forget about this. Skip is: Copier, if this file exists in the destination, forget about it.

So, your use case is skip.

README.md
sbidoul: @yajo At least in the released version of copier that I pipx installed, `--skip` does not take arguments. Is it something new?

Yajo: Humm sorry I was mistaken :speak_no_evil: 

I was confusing it with the `skip_if_exists` option, which is only available through API right now.

Gonna open some issues and work on that.

Yajo: Opened https://github.com/pykong/copier/issues/214 and https://github.com/pykong/copier/issues/215

Yajo:
AFAICS, if you want to automate this stuff, the best you can do is use python itself:

```py
from copier import copy
copy(src, dst, force=True, ...)
```

I think we can merge this. This could evolve progressively, but this seems good enough as a starting point.

WDYT @sbidoul?

Yajo:
OK, squashing then.

I think all related changes are in MQT.... :thinking:

PR #3, opened 2020-06-03
Add placeholder for exclude list of uninstallable addons

sbidoul:


sbidoul:
Ok, I think this is the last one before create the 14.0 branch in all repos.

It is to support repos that want some addons with installable=False. This disables linting on them, while ensuring linting is reenabled automatically when installable becomes True again.

Here is an example : https://github.com/OCA/mis-builder/commit/38f75413687e79867840c9e6f86ee207394fed86#diff-82c3b36709926de3a1b27bea0d8258cc

pedrobaeza (approved):

lmignon (approved):

PR #8, opened 2020-09-23
Prepare pre-commit-config for odoo v14

Yajo:
- Move linters after formatters. It makes no sense to lint code that is going to be reformatted.
- Store revisions in variables and use them.
- Where there are more changes than just the revisions, add needed logic.
- Update dependencies.
- Add v14 to CI.

@Tecnativa TT22772


joao-p-marques (approved):

.pre-commit-config.yaml.jinja
```diff
@@ -13,10 +40,23 @@ exclude: |
...
-    rev: 19.10b0
+    rev: {{ repo_rev.black }}
     hooks:
       - id: black
+
+  {#- Old prettier required a separate step for the XML plugin because it
+      didn't support including plugins #}
+  {%- if odoo_version < 14 %}
   - repo: https://github.com/prettier/prettier
     rev: "1.19.1"
```
lmignon, line 61: @Yajo Why don't you use parameters `repo_rev.prettier` and `repo_rev.prettier_xml` also for version < 14? It's nice to see at the same place the versions of all the plugins used.  (Same applies to pylint and pylint-odoo)

.pre-commit-config.yaml.jinja
```diff
@@ -1,3 +1,30 @@
+{#- This namespace holds repo versions; other variable definitions are found
```
lmignon, line 1: @Yajo Would it be more readable to duplicate the full definition of the pre-commit hooks? Personally, I find that the conditional inline addition/removal of hooks is not very readable. :thinking:

.pre-commit-config.yaml.jinja
Yajo: Yes, no problem.

.pre-commit-config.yaml.jinja
Yajo: I have encountered feelings about that.

On one side, indeed the configs would be more readable.

On the other side, maintenance is easier. For example,  #9 would be added in a single file, and both v13 and v14 repos would benefit from that enhancement.

Although reading all those "ifs", in reality I can see that from the time we delivered v13 branches, the main diff is that all projects we use started supporting pre-commit out of the box (except pylint-odoo with https://github.com/OCA/pylint-odoo/issues/296, which would be nice if fixed before v14 creation so we can remove that hack from here). So, we can expect less diffs the next year; just the versions probably.

Then, maybe I can create one for v13 only, and other one for any other version. If for v15 we see a lot of diff, we do the same. Otherwise, we just update versions. WDYT?

sbidoul: My instinct would also have been to create a template branch for 13.0, 14.0, etc but I guess that ship has sailed, so let's see where it goes.

.pre-commit-config.yaml.jinja
Yajo: Finally I have not implemented this because it doesn't make sense with the other suggestion that I followed.

.pre-commit-config.yaml.jinja
Yajo: Kudos for not following your instinct! :blush: 

We have a templating engine for some reason. :wink: 

I have applied your suggestions. I hope you find it more readable now.

Yajo:
One little thing regarding the pre-commit stuff: if there are plans for fixing https://github.com/OCA/oca-github-bot/issues/55 some day, different pre-commit hook versions across module branches would dificult that a lot.

For example, code formatted with black 19.10b0 and 20.8b1 is quite different, so doing automatic fw/bw ports becomes much harder.

So, maybe we need to reconsider the decision of freezing hooks per odoo version, and, instead, update all of them at once, regardless of odoo version (in this context, this means branches 13.0 and later only).

The obvious problem of that decision would be that some open PRs that were green before the pre-commit update could become :x: once merged. That has 2 possible solutions: 

1. Make the merge bot apply pre-commit if needed.
1. Ask the OP to rebase and update the code (assuming some will get angry).

This PR is currently implementing the current decision of pre-commit hook per odoo version, and I'm not saying it's the right moment to change that. It's not a decision that we have to take right now because https://github.com/OCA/oca-github-bot/issues/55 is far from fixed, but I'm just commenting it because we should keep this in mind.

Yajo:
This should be ready to squash and merge. May you review again please @lmignon and @sbidoul?

.pre-commit-config.yaml.jinja
```diff
@@ -1,3 +1,28 @@
...
+  {#- HACK https://github.com/OCA/pylint-odoo/issues/296#issuecomment-700317722 #}
+  {%- set repo_rev.pylint = "pylint-2.5.3" %}
+  {%- set repo_rev.pylint_odoo = "3.5.0" %}
+  {%- set repo_rev.pyupgrade = "v2.7.2" %}
+  {%- set repo_rev.setuptools_odoo = "2.5.9" %}
+{%- endif %}
+
+{#- Older versions that differ a lot have their own hardcoded templates for readability #}
+{%- if odoo_version < 14 %}
+  {%- include "version-specific/%s/.pre-commit-config.yaml" % odoo_version %}
```
lmignon, line 22: :heart:

lmignon (approved):
Thank you @Yajo for all this hard work.

Yajo:
Thank you too, merging then. :blush:

PR #9, opened 2020-09-24
IMP: apply autoflake to remove unsed imports

celm1990:
continuation of https://github.com/OCA/maintainer-quality-tools/pull/676

CC @Yajo @sbidoul

### DIFF
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -2,6 +2,7 @@
 {%- set repo_rev = namespace() %}
 
 {%- if odoo_version >= 14 %}
+  {%- set repo_rev.autoflake = "v1.4" %}
   {%- set repo_rev.black = "20.8b1" %}
   {%- set repo_rev.eslint = "v7.8.1" %}
   {%- set repo_rev.flake8 = "3.8.3" %}
@@ -47,6 +48,11 @@ repos:
         entry: found forbidden files; remove them
         language: fail
         files: "\\.rej$"
+  - repo: https://github.com/myint/autoflake
+    rev: {{ repo_rev.autoflake }}
+    hooks:
+      - id: autoflake
+        args: ["-i", "--ignore-init-module-imports"]
   - repo: https://github.com/psf/black
     rev: {{ repo_rev.black }}
     hooks:
--- a/version-specific/13.0/.pre-commit-config.yaml
+++ b/version-specific/13.0/.pre-commit-config.yaml
@@ -13,6 +13,11 @@ exclude: |
 default_language_version:
   python: python3
 repos:
+  - repo: https://github.com/myint/autoflake
+    rev: v1.4
+    hooks:
+      - id: autoflake
+        args: ["-i", "--ignore-init-module-imports"]
   - repo: https://github.com/psf/black
     rev: 19.10b0
     hooks:


legalsylvain (approved):

joao-p-marques (approved):

.pre-commit-config.yaml
```diff
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/myint/autoflake
+    rev: master
```
sbidoul, line 3: We should never reference a branch in pre-commit hooks, otherwise it defeats the pre-commit caching mechanism. It must be a tag or commit.

sbidoul (changes requested):

.pre-commit-config.yaml
celm1990: done, lasted version are used https://github.com/myint/autoflake/releases/tag/v1.4

Yajo (approved):
I like it.

celm1990:
@sbidoul changes are done, any other suggestion or can be merge this PR?

sbidoul:
I've never used autoflake myself nor seen the need to use it so a few questions that come to mind:

My main question is: does this risk making green code red (by causing a diff on unused imports that would not have been considered an error by another linter) ? If someone really wants to do an import for some (good or bad reason) and flags it with # noqa`, will autoflake leave it alone ?

joao-p-marques:
From the `autoflake` documentation (https://github.com/myint/autoflake#excluding-specific-lines), seems like that is what would happen @sbidoul

Yajo:
I've been testing it recently and it worked pretty well, but I can't say for every corner case. In theory, it's just going to autofix some problems where flake8 would set the branch red anyways, so it shouldn't make fail anything that weren't failing before.

Yajo:
After merging #8, this has conflicts. Plase fix them, so we can use this in the next release, which I hope happens today or tomorrow.

celm1990:
ping @sbidoul @Yajo

.pre-commit-config.yaml
```diff
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/myint/autoflake
+    rev: v1.4
+    hooks:
+      - id: autoflake
+        args: ["-i", "--remove-all-unused-imports", "--ignore-init-module-imports"]
```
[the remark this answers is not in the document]
celm1990: i only removed arg `--remove-all-unused-imports` is implicit, but other args is need to be explicit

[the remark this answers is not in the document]
sbidoul: Oh sorry I was not clear, it is the whole hook that is not necessary in this file. Only in `.pre-commit-config.yaml.jinja` and in `version-specific/13.0`.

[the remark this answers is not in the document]
celm1990: is ok now??

.pre-commit-config.yaml.jinja
```diff
@@ -47,6 +47,11 @@ repos:
         entry: found forbidden files; remove them
         language: fail
         files: "\\.rej$"
+  - repo: https://github.com/myint/autoflake
+    rev: v1.4
```
Yajo, line 51: Please use a version variable (you have to define it also around line 5):

```suggestion
    rev: {{ repo_rev.autoflake }}
```

Yajo (changes requested):
Just one little change to go, thanks!

.pre-commit-config.yaml.jinja
celm1990: change done

sbidoul:
Thanks!

PR #17, opened 2020-10-07
Set language_version for prettier

### LINKED ISSUE #15: prettier requires at least version 10.13.0 of Node, please upgrade
Example [here](https://travis-ci.com/github/OCA/mis-builder/jobs/396016494#L329-L330).

Does @Yajo or anyone know the fix for that?

sbidoul:
The default available on travis is too old.

fixes #15

### DIFF
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -38,6 +38,7 @@ exclude: |
   (LICENSE.*|COPYING.*)
 default_language_version:
   python: python3
+  node: "14.13.0"
 repos:
   - repo: local
     hooks:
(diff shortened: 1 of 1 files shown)


Yajo:
Cool! I was trying to help @ernestotejeda to fix this the other day and didn't have a clue...

Do you think it'd make more sense to update the `default_language_version` instead for `node`, just like we have for `python`?

sbidoul:
> Do you think it'd make more sense to update the default_language_version instead for node, just like we have for python?

Ok, I'll try that. It's probably more coherent to have the same node version for all hooks that need it.

Yajo:
Yes, that's what I thought.

sbidoul:
Gah, now tests fails here on poetry install...

Yajo:
Possibly a cache problem. I need admin access to this repo to fix it.

sbidoul:
@Yajo I just gave you admin access.

Yajo:
Note: to apply this workaround: https://github.com/actions/cache/issues/2#issuecomment-673493515

sbidoul:
@yajo while you are at it you may want to look into #18

Yajo:
After merging #19, if you rebase this will probably get fixed.

Otherwise, we can alter the `$CACHE_DATE` secret later, to force cache "clearing", and again it should get fixed.

sbidoul:
Ok, green. Almost there.

PR #20, opened 2020-10-08
Remove pylint-odoo missing-import-error and missing-manifest-dependency checks

sbidoul:
- `missing-import-error` is [not required for Odoo >= 12](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#importerror)
- To the best of my knowledge, `missing-manifest-dependency` does not work correctly when the external dependency is a PyPI distribution name that is different from the import name, as pylint-odoo has no way to correlate them. Since Odoo 13, PyPI distribution names are recommended in manifest external dependencies for ease of discovery when they differ from the import name.


.pylintrc-mandatory.jinja
```diff
@@ -38,8 +38,6 @@ enable=anomalous-backslash-in-string,
     method-inverse,
     method-required-super,
     method-search,
-    missing-import-error,
-    missing-manifest-dependency,
```
Yajo, line 42: Maybe we could add `missing-manifest-dependency` to the optional linters?

sbidoul: Yep, I was considering that too. Done.

PR #24, opened 2020-10-18
Fixes #21

### LINKED ISSUE #21: oca-update-pre-commit-excluded-addons not running on addon flipping from installable false to true
https://github.com/OCA/maintainer-tools/pull/474

sbidoul:


sbidoul:
@Yajo looks like poetry install is acting up again ?

.pre-commit-config.yaml.jinja
```diff
@@ -52,7 +52,7 @@ repos:
         language: fail
         files: "\\.rej$"
   - repo: https://github.com/oca/maintainer-tools
-    rev: 2c66e72
+    rev: fd75d80
```
Yajo, line 55: Wouldn't it be better to do tag releases like everyone else? :thinking: 

That would allow autoupdating it.

Yajo:
> looks like poetry install is acting up again ?

Let me check that

Yajo:
Please rebase now that #28 is merged.

Yajo (approved):

.pre-commit-config.yaml.jinja
sbidoul: We could do maintainer-tools tag, yes.

PR #27, opened 2020-10-18
[ADD] add development-status-allowed check

sbidoul:
It looks like this check was missing.

@moylop260 is this correct ?


Yajo:
Please rebase now that #28 is merged.

Yajo (approved):

sbidoul:
@moylop260 should we add this check to the default mandatory checks in MQT (for branches that don't use pre-commit).

moylop260:
> @moylop260 should we add this check to the default mandatory checks in MQT (for branches that don't use pre-commit).

👍

PR #28, opened 2020-10-21
Update test workflow

Yajo:
- Use poetry action, easier.
- Cache only venv.
- Fix bug in cache key creation.


Yajo:
Fast-tracking as this fixes the branch and eases its maintenance.

Issue #22, opened 2020-10-18
Test for python 3.6 and 3.8 ?


sbidoul:

Yajo:
I guess the best would be a matrix of all supported python versions, but otherwise I'd pick up the lowest and highest, which are 3.6 and 3.9 ATM.

joao-p-marques:
Could this be easier if the test stage was moved to Github Actions, as with the `pre-commit` job? It has great support for that matrix approach with parallel runners.

Yajo:
I'd love that, but MQT is quite biased towards Travis and that'd mean a lot of work I guess. And travis also supports job matrix.

PR #29, opened 2020-10-21
Test python 3.6-3.8

yajo:
Python 3.9 will only land in Copier v6 as seen in https://github.com/copier-org/copier/pull/294.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        python-version:
+          - 3.6
+          - 3.7
+          - 3.8
         odoo-version:
           - 13.0
           - 14.0
@@ -23,6 +27,8 @@ jobs:
       - uses: actions/checkout@v2.3.3
       - name: Install python
         uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
       - name: Python Poetry Action
         uses: abatilo/actions-poetry@v2.1.0
       - name: generate cache key PY
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,6 @@ jobs:
       matrix:
         python-version:
           - 3.6
-          - 3.7
           - 3.8
         odoo-version:
           - 13.0


sbidoul (dismissed):
@Yajo #22 was rather intended to update `.travis.yml.jinja` if/when an agreement is found in the mailing list discussion. But otherwise :+1: for testing the template with different python versions.

Yajo:
Lol, sorry!

Well, this is good anyway IMHO.

sbidoul:
@yajo there is one check (`test (13)`) that stays pending.

bealdav:
Thanks a lot to @Yajo and @sbidoul for your agility and reactivity for oca tool box.
The improvements of workflows for maintainability are really big, congratulations for this investments. It's incredible. I told a word about it in mailing list as many people.
I see you make the impossible to satisfy everybody, those which (aka casse pied) want to use recent python version and others with as many versions as Odoo propose. For sure that's the best intentions.
But as always I reach discussions probably too often but I have a comment also here. We know that github is generous with open source projects which are financed by proprietary projects. But support all these versions have consequence, I suppose, on the earth environment. Yes we can make as many CI workflows as we want, but finally it increase some major problems as everybody 'll meet in its future life. I'm sure many oca contributors also are citizen implied for earth future in misc actions. On my side I try to do some actions but really minor ones at least for their impact. Sometimes I ask myself what can do with numeric/digital activities. Reduce impact is difficult without additional impact on Odoo growth.
Finally it's difficult to find real actions. Actions could lead to have less comfort with odoo landscape. And there is a question. Does it worth do it that. What if I hadn't this. I think a large people majority in oca think about ci-check a lot of python versions. Does it worth to do it ? What's the impact only odoo had update their setup file to 3.7 (they maybe forgot it), which gain for environment ? Completeness position to be fully aligned on odoo or a position less generous. Which concession we could accept, I don't know. Finally what could we do ? What's the best option for all ?
I'll let decide. Again, thanks a lot for your hard work here.

Yajo:
@sbidoul sorry that was configured as a required check. I updated those, it's green now.

@bealdav I like the subject you raise. It's indeed something not very often thought. I believe it would have more impact in other OCA repos where most people are very active. This one has big chances of diversity, as it will be executed on many devs laptops, which have very different environments. Also this repo is getting a lot of traction lately, but won't be the case for most of the year, at least until a new odoo release approaches. I think the balance is somehow justified here, but I wouldn't reject to roll back this if others think the contrary.

Maybe we could just skip the oldest one, 3.6?

sbidoul:
If we want to save build time (which we generally do for environmental and practical reasons), I think the best is test the oldest and newest python version we want to support (could be 3.6 and 3.9 in this case). And trust python to keep compatibility in between.

Yajo:
Nice, changed. 3.9 is not yet supported by copier, so this tests 3.6 and 3.8 now.

PR #31, opened 2020-10-21
Add setuptools-odoo-get-requirements

### LINKED ISSUE #23: Pre-commit hook to generate requirements.txt
See https://github.com/acsone/setuptools-odoo/pull/42

sbidoul:
Generate `requirements.txt` from manifest python `external_dependencies` and `setup.py` overrides if any.

Closes #23

### DIFF
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -15,7 +15,7 @@
   {%- set repo_rev.pylint = "pylint-2.5.3" %}
   {%- set repo_rev.pylint_odoo = "3.5.0" %}
   {%- set repo_rev.pyupgrade = "v2.7.2" %}
-  {%- set repo_rev.setuptools_odoo = "2.5.10" %}
+  {%- set repo_rev.setuptools_odoo = "2.6.0" %}
 {%- endif %}
 
 {#- Older versions that differ a lot have their own hardcoded templates for readability #}
@@ -122,6 +122,13 @@ repos:
     rev: {{ repo_rev.setuptools_odoo }}
     hooks:
       - id: setuptools-odoo-make-default
+      - id: setuptools-odoo-get-requirements
+        args: [
+          "--output",
+          "requirements.txt",
+          "--header",
+          "# generated from manifests external_dependencies"
+        ]
   - repo: https://gitlab.com/PyCQA/flake8
     rev: {{ repo_rev.flake8 }}
     hooks:
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -122,12 +122,14 @@ repos:
     rev: {{ repo_rev.setuptools_odoo }}
     hooks:
       - id: setuptools-odoo-make-default
+      {%- if generate_requirements_txt %}
       - id: setuptools-odoo-get-requirements
         args:
           - --output
           - requirements.txt
           - --header
           - "# generated from manifests external_dependencies"
+      {%- endif %}
   - repo: https://gitlab.com/PyCQA/flake8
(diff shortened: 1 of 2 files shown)


sbidoul:
And here is a demo of what it does: https://github.com/OCA/l10n-france/pull/249/files

rvalyi:
looks good. However just a dumb question: if the repo requirements.txt is incomplete, Travis tests for the module will fail anyway forcing the author to complete it, right? So what is the benefit getting it generated?

sbidoul:
Hi @rvalyi,

> So what is the benefit getting it generated?

The main reason is that we want to ensure that the manifests are correct wrt external dependencies, for consistency and a better experience for users installing OCA addons with pip.

Also, in `MQT_DEP=PIP` mode `requirements.txt` is not used for testing (see [the thread](https://odoo-community.org/groups/contributors-15/contributors-174762?mode=thread&date_begin=&date_end=)) because we install dependencies at the individual addon granularity.

sbidoul:
So in short, remove a source of redundancy.

rvalyi:
makes sense, thanks for the explanations.

.pre-commit-config.yaml.jinja
```diff
@@ -122,6 +122,13 @@ repos:
     rev: {{ repo_rev.setuptools_odoo }}
     hooks:
       - id: setuptools-odoo-make-default
+      - id: setuptools-odoo-get-requirements
```
Yajo, line 125: I think this should be wrapped by an `{% if dependency_installation_mode == 'PIP' %}` block, right?

.pre-commit-config.yaml.jinja
```diff
@@ -122,6 +122,13 @@ repos:
...
+        args: [
+          "--output",
+          "requirements.txt",
+          "--header",
+          "# generated from manifests external_dependencies"
+        ]
```
Yajo, line 131: Please don't:

```suggestion
        args:
          - --output
          - requirements.txt
          - --header
          - "# generated from manifests external_dependencies"
```

We don't use JSON for a reason :grin:

.pre-commit-config.yaml.jinja
sbidoul: @Yajo I don't think so: this just ensures that `requirements.txt` is consistent with the declarations in manifests (and possible overrides in setup.py), which must be the case in both installation modes. 

Generating `oca_dependencies.txt` (#32) will be for `MQT_DEP=PIP` only, of course.

.pre-commit-config.yaml.jinja
sbidoul: If you prefer. But that's nitpicking. It's valid yaml :)

.pre-commit-config.yaml.jinja
sbidoul: Although generating requirements.txt is useful regardless of the installation mode, we don't want to activate it by default on 13.0 branches since it could make repos with imperfect manifests red. So I added a question to control this. @Yajo

sbidoul:
@yajo ok with this one? I'm waiting for it before pushing the updates to 14.0 branches.

copier.yml
```diff
@@ -76,6 +76,12 @@ dependency_installation_mode:
   help:
     Choose how to install module dependencies. Right now, "PIP" mode is experimental.
 
+generate_requirements_txt:
+  default: yes
+  help:
```
Yajo, line 81:
```suggestion
  type: bool
  help:
```

Yajo (changes requested):

Yajo:
Just that tiny change and ready to  go!

PR #35, opened 2020-10-26
Use strict mode in prettier xml plugin

### LINKED ISSUE #34: [14.0] prettier-xml: A little bit too invasive
I had issue with a report : https://travis-ci.com/github/OCA/event/jobs/406980086#L369

prettier-xml is adding carriage return to xml values. In that case, this is annoying as Odoo does not trim, so when installing, the reference to the report is not found.

Values shouldn't be modified.

Yajo:
This will make XML files be less pretty, but also less surprising.

Fix #34.

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -5,4 +5,3 @@ printWidth: 88
 proseWrap: always
 semi: true
 trailingComma: "es5"
-xmlWhitespaceSensitivity: "ignore"


pedrobaeza (approved):

pedrobaeza:
This should be propagate to 14.0 branches. It's better to do it with `[ci skip]` for avoiding a lot of Travis builds.

sbidoul:
> This should be propagate to 14.0 branches. It's better to do it with [ci skip] for avoiding a lot of Travis builds.

Yes, waiting for #31 then I'll do the push, with ci skip indeed.

PR #36, opened 2020-10-26
Pin node version for 13.0 too

sbidoul:


sbidoul:
I'm not going to use copier to update the 13.0 branches but let's record this anyway.

pedrobaeza (approved):

joao-p-marques (approved):

sbidoul:
This patch has been deployed on the 13.0 branches.

PR #47, opened 2021-01-03
[FIX][14.0] only run prettier xml on .xml files (like 13.0)

nilshamerlinck:
In the continuity of https://github.com/OCA/oca-addons-repo-template/pull/46

### DIFF
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -72,6 +72,9 @@ repos:
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v{{ repo_rev.prettier }}
     hooks:
+      # first prettier hook without plugin-xml, on all supported files
+      - id: prettier
+      # second prettier hook with plugin-xml, on xml files only
       - id: prettier
         name: prettier + plugin-xml
         additional_dependencies:
@@ -79,6 +82,7 @@ repos:
           - "@prettier/plugin-xml@{{ repo_rev.prettier_xml }}"
         args:
           - --plugin=@prettier/plugin-xml
+        files: \.xml$
   - repo: https://github.com/pre-commit/mirrors-eslint
     rev: {{ repo_rev.eslint }}
     hooks:


guewen (dismissed):

.pre-commit-config.yaml.jinja
```diff
@@ -79,6 +79,7 @@ repos:
           - "@prettier/plugin-xml@{{ repo_rev.prettier_xml }}"
         args:
           - --plugin=@prettier/plugin-xml
+        files: \.xml$
```
joao-p-marques, line 82: I'm not sure, but shouldn't this be something like this?
```suggestion
        files: "\\.xml$"
```
This is following the same as in https://github.com/OCA/oca-addons-repo-template/pull/46/files#diff-643aa373eb3374e32d8b29a56a7d59d32f2f6bc86fcb5efd1d926f51b1364f17R29

nilshamerlinck: Hi João
- on a yaml point of view `\.xml$` and `"\\.xml$"` are equivalent 
- jinja template evaluation doesn't mess with `\.xml$` so I don't see the need for what I perceive as the more complex form  `"\\.xml$"`
- in the end it sounds like a subjective preference; I don't think that the yaml parser of prettier supports enforcing one or the other form (see https://github.com/prettier/prettier/issues/5350) but you might want to create an issue at this repo level if you think it's worth :-)

joao-p-marques: Ok, I was just suggesting for a matter of consistency, but your approach seems perfectly fine :smiley: 
I thought the escaping character `\` was needed but indeed it isn't :sweat_smile: 
Thank you for the explanation :+1:

guewen (dismissed):

pedrobaeza:
@sbidoul is it OK for you? I suppose that this needs to be re-applied on all 14.0 branches.

sbidoul:
Why do we want to do this ? Will that not make prettier run on xml files *only*, and disable it for js files ?

In 13, there is a subtle difference: there are *two* prettiers hooks: the default one that runs on many extensions except xml, and one that runs on xml only.

joao-p-marques:
>Will that not make prettier run on xml files only, and disable it for js files?

Looking at it now I think @sbidoul is right. If the point is to enable the prettier xml plugin only for xml files, maybe the solution is to create 2 separate hooks: 1 without the xml plugin that will apply to all files, and another that will have the spec suggested in this PR. 

Could that work?

pedrobaeza:
And what is the current problem?

nilshamerlinck:
I missed the double prettier hook indeed.

The problem I faced initially was that `prettier/plugin-xml` fails to parse some xsd files, e.g. https://github.com/OCA/edi/blob/18ff921968ed79cd745403183cc040b815c19dab/base_ubl/data/xsd-2.1/common/UBL-xmldsig-core-schema-2.1.xsd

I don't think that we should run `prettier/plugin-xml` on anything else than odoo xml files as [other xml types of files](https://github.com/prettier/plugin-xml/blob/master/src/languages.js) have high chance of being provided by third parties.

pedrobaeza:
I think the problem is the path of your XSD file. Why is inside `data` folder? That folder is for Odoo XML data files.

sbidoul:
I see. My stance on this is that we should list explicitly the file extensions we want prettier to handle.

pedrobaeza:
Wasn't certain folders blacklisted for prettier (like /static/lib)?

nilshamerlinck:
In short term, changed to match with the behavior of 13.0, as described by @joao-p-marques :

> create 2 separate hooks: 1 without the xml plugin that will apply to all files, and another that will have the spec suggested in this PR.

Is it ok?

joao-p-marques (approved):
LGTM :+1:

Yajo (approved):

pedrobaeza:
@sbidoul is it correct now?

Yajo:
Oh sorry I now understood https://github.com/OCA/oca-addons-repo-template/pull/47#issuecomment-754544689 and how it relates to the code. I think this is not the solution he was suggesting.

nilshamerlinck:
Hi @sbidoul is https://github.com/OCA/oca-addons-repo-template/pull/48 what you had in mind?

nilshamerlinck:
@pedrobaeza while doing some more research on this topic, I stumbled on this comment: https://github.com/prettier/plugin-xml/issues/17#issuecomment-555445966

`/web/static` being excluded indeed, in an early version of the hook focused on xml files only

Do you think it would still be needed in some repos? Or should we better use explicit `.prettierignore` files for such cases?

pedrobaeza:
I think that adding that sane defaults is good, but @Yajo and @sbidoul are the closest ones to the final implementation to say.

nilshamerlinck:
well, actually all good @pedrobaeza I just noticed that [`/static/(src/)?lib/` is already globally excluded by default](https://github.com/OCA/oca-addons-repo-template/blob/master/.pre-commit-config.yaml.jinja#L38) :)

pedrobaeza:
Great then :)

PR #48, opened 2021-01-07
[IMP][14.0] run prettier on an explicit list of extensions

nilshamerlinck:


joao-p-marques (commented):
Would it still make sense to split it into 2 hooks (one w/ the xml plugin and another without it) to avoid future problems?

sbidoul (approved):

nilshamerlinck:
@joao-p-marques Do you remember what were the problems that lead to introducing 2 hooks? I did some research but could not find them :/

sbidoul:
I think we needed two hooks because older versions of the xml plugin did not add the xml extention to the list of files supported by the default hook. Not sure. But if this version actually does something on all extensions (has someone tested it?), one hook should be fine.

joao-p-marques:
@nilshamerlinck my suggestion came from the fact that the xml plugin was causing problems with some file types, as stated in https://github.com/OCA/oca-addons-repo-template/pull/47. Obviously, this solves it with another approach, but maybe leaving the 2 hooks separate can be a safe precaution to avoid more problems with the posibility that the xml plugin wrongly parses other file types.

joao-p-marques:
And, as @sbidoul said, could also allow us to pin different versions for each hook.

nilshamerlinck:
ok @joao-p-marques but my initial problem was a bit different, as it was due to the xml plugin failing on a xsd file, that would not have been parsed by prettier itself.

If we need to pin a different version in the future, we can do it at that moment, with a specific commit ;-)

@sbidoul yes, I tried it, it runs on each concerned files as expected :)

sbidoul:
Cool. If @Yajo has no objections I can deploy this.

.pre-commit-config.yaml.jinja
```diff
@@ -72,17 +72,14 @@ repos:
...
-      - id: prettier
-        name: prettier + plugin-xml
+        name: prettier (with plugin-xml)
         additional_dependencies:
           - "prettier@{{ repo_rev.prettier }}"
           - "@prettier/plugin-xml@{{ repo_rev.prettier_xml }}"
         args:
           - --plugin=@prettier/plugin-xml
-        files: \.xml$
+        files: \.(js|xml|css|less|scss)$
```
Yajo, line 82: There are a few more that are safe to reformat. Some of them are actually used commonly; others maybe will not be used, but in case any of these lands into our codebase, I think it will be nice to keep them autoformatted. I sorted the list, BTW:

```suggestion
        files: \.(css|htm|html|js|json|jsx|less|md|scss|toml|ts|xml|yaml|yml)$
```

Yajo (changes requested):
Indeed, it was split in 2 hooks before due to prettier limitations that went away at some release of prettier and/or the XML plugin.

I also wonder if the problem from https://github.com/OCA/oca-addons-repo-template/pull/47#issuecomment-754543484 wouldn't be easier to fix with just some prettier-ignore comments.

Well, never mind, this patch doesn't seem problematic anyway if completed.

.pre-commit-config.yaml.jinja
nilshamerlinck: updated

Yajo (approved):

sbidoul (approved):

Yajo:
@joao-p-marques your veredict?

joao-p-marques:
Well, I still think having 2 separate hooks could help avoid problems in the future, as @sbidoul said 
> I think we needed two hooks because older versions of the xml plugin did not add the xml extention to the list of files supported by the default hook.

If someday we find any incompatibility between versions or some more errors in formatting due to the plugin, it would be easier to fix, or never happen in the first place.

However, if that is not causing any problems now, that shouldn't be blocking :smiley:

joao-p-marques (approved):

Yajo:
Yes, that's no problem.

FTR https://github.com/OCA/oca-addons-repo-template/pull/48#issuecomment-756174006 is true. That problem no longer exists.

sbidoul:
I have tagged v1.1.2. I don't think I'm going to deploy it to all repos. PSC affected by this issue can run `copier -f update`.

PR #49, opened 2021-01-15
[RFR] pre-commit: prevent pyupgrade from replacing printf with str.format()

StefanRijnhart:
So as not to encourage developers to use str.format() as there are cases where it can introduce security issues.

See https://github.com/OCA/pylint-odoo/issues/302

### DIFF
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -114,6 +114,7 @@ repos:
     rev: {{ repo_rev.pyupgrade }}
     hooks:
       - id: pyupgrade
+        args: ["--keep-percent-format"]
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

moylop260 (approved):
👍

joao-p-marques (approved):

Yajo (approved):

StefanRijnhart:
I still encounter this behaviour, for instance in reporting-engine/14.0 where the config file is from the last batch update of 2020-12-11 (https://github.com/OCA/reporting-engine/commit/69da04e57df32e1dbfcfb10457541ec188413bdf). Can a new update be triggered or is it too much work?

PR #51, opened 2021-01-27
[FIX] prettier for v13 broken in mirror

Yajo:
https://github.com/pre-commit/mirrors-prettier/issues/6 affects our repos, as it can be seen on https://github.com/OCA/social/runs/1770760744?check_suite_focus=true#step:4:110 for https://github.com/OCA/social/pull/664 and it produces random results when running `pre-commit` in different environments.

I apply here the workaround explained in that issue.

At the same time, I double-checked that we can run all prettier jobs in a single hook, so I take this chance to apply that logic, and to limit by file extensions, to make it work just like the v14 branch with #48, more predictable and efficent.

### DIFF
--- a/version-specific/13.0/.pre-commit-config.yaml
+++ b/version-specific/13.0/.pre-commit-config.yaml
@@ -47,12 +47,12 @@ repos:
     rev: v1.19.1
     hooks:
       - id: prettier
-      - id: prettier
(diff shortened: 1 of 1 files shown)


Yajo:
I'm sorry but I need to fast-track this one. 🙈 

/ocabot merge nobump

PR #53, opened 2021-03-19
Increase line length

### LINKED ISSUE #52: [Question] Is .flake8 line length used?
The black configuration makes it adjust to a line legth of 88, but looking at the [configyuration of flake8](https://github.com/OCA/oca-addons-repo-template/blob/master/.flake8#L2) I see an 80.

This is at least misleading if it is not really used. Can this be removed or clarified?

Yajo:
Fix #52 by adopting the new official recommendation from https://github.com/psf/black/pull/2050. Now the linter and the formatted are configured likewise.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length = 80
+max-line-length = 88
 max-complexity = 16
 # B = bugbear
 # B9 = bugbear opinionated (incl line length)


pedrobaeza:
I think this can't be done in a stable branch, but on the upcoming one.

Yajo:
We're not reducing limits, but increasing them, so chances to break something are 0%. There's no practical reason for that.

pedrobaeza (approved):
OK, no special problem then

joao-p-marques (approved):

PR #55, opened 2021-04-12
Improve isort compatibility with black

tmotyl:
Add ensure_newline_before_comments, see:
https://black.readthedocs.io/en/stable/compatible_configs.html#configuration

To trigger the rule, add a comment before some import.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -10,3 +10,4 @@ known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
 default_section=THIRDPARTY
+ensure_newline_before_comments = True


pedrobaeza (approved):

matteoopenf (approved):

.isort.cfg
```diff
@@ -10,3 +10,4 @@ known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
 default_section=THIRDPARTY
+ensure_newline_before_comments = True
```
Yajo, line 13: Possibly we should just dump most options here and add:

```
profile = "black"
```

See https://pycqa.github.io/isort/docs/configuration/black_compatibility/

(Note: in v13 branches, we used older isort, which didn't have that option).

Yajo:
Well, in any case, it's an enhancement. 🤷🏼‍♂️

PR #58, opened 2021-05-04
[IMP] autoflake more

Yajo:
This should let us rely more on the cleaning power of autoflake. It removes all unused imports (not just from the stdlib), variables and dict keys. Also, it expands star imports (which shouldn't happen usually, but in any case it helps).

### DIFF
--- a/.pre-commit-config.yaml.jinja
+++ b/.pre-commit-config.yaml.jinja
@@ -64,7 +64,13 @@ repos:
     rev: {{ repo_rev.autoflake }}
     hooks:
       - id: autoflake
-        args: ["-i", "--ignore-init-module-imports"]
+        args:
+          - --expand-star-imports
+          - --ignore-init-module-imports
+          - --in-place
+          - --remove-all-unused-imports
+          - --remove-duplicate-keys
+          - --remove-unused-variables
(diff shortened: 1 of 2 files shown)


joao-p-marques (commented):
:+1: 

Do you have any example of the result in a current project?

pedrobaeza:
Is this safe to apply in any existing repo?

Yajo:
It should be safe because it follows the same rules as flake8 itself:

- `--expand-star-imports` fixes F403
- `--remove-all-unused-imports` fixes F401
- `--remove-duplicate-keys` fixes F601
- `--remove-unused-variables` fixes F841

And according to our `.flake8` all "F" codes are selected: https://github.com/OCA/oca-addons-repo-template/blob/761a8c274b513227dd5151bfa41f41cbc8bdaefe/.flake8#L6

[See flake8 error codes cheat sheet here](https://flake8.pycqa.org/en/latest/user/error-codes.html).

So, the only difference is that for those specific errors, pre-commit would fix them automatically instead of just complaining about them.

There's no example to show in OCA code base because flake8 was already forbidding them.

joao-p-marques (approved):

pedrobaeza (approved):

PR #59, opened 2021-05-14
remove oca_dependencies.txt from template

sbidoul:
The last change (in #57) is causing copier update merge conflicts
on all repos that touched it, essentially preventing a mass update to
the latest template.

Since this file will either go away or be auto generated, it has little value in the template anyway.

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,1 +1,0 @@
-# See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca-dependencies-txt


sbidoul:
Fast tracking, so I can launch the mass update while I've got the motivation to do it :) We can still revert if folks disagree.

legalsylvain:
:+1:

Yajo:
It could have been added to `_skip_if_exists` instead, but it's indeed not the most useful file anyway if bots add it.
https://copier.readthedocs.io/en/stable/configuring/#skip_if_exists

PR #60, opened 2021-06-01
propose global owa

nosinga:
Hi,

I like to propose the following change

globals:
..  
owl: readonly
..

In this way it is not necessary anymore to use the following hint in the script

/* global owl */


StefanRijnhart:
Thanks! Can you reword the commit to include `eslint` and `config` (e.g. `eslint config: allow global "owl"`)?

PR #61, opened 2021-06-08
Test with GitHub actions

sbidoul:
As discussed on the contributors ML.

Before we merge this, I'll first deploy an update to the merge bot that let it ignore some GitHub checks.
This will allow for a transition phase where we can see how this behaves without blocking merges in case of problem.

Examples of this at work:
- https://github.com/oca/mis-builder 14.0
- https://github.com/OCA/product-variant/pull/218
- https://github.com/OCA/sale-workflow/pull/1607

Things that may need fixing of existing repos when we deploy this:

- development status (e.g. a Beta module that depends on an Alpha module)
- license compatibility (e.g. a LGPL module that depends on an AGPL module)
- missing external_dependencies (unlikely in 14.0)
- adding apt packages in a `deb` section of `external_dependencies`


.github/workflows/test.yml.jinja
```diff
@@ -0,0 +1,65 @@
...
+        run: manifestoo -d . check-licenses
+      - name: development status check
+        run: manifestoo -d . check-dev-status --default-dev-status=Beta
+      - run: oca_run_tests
+      - uses: codecov/codecov-action@v1
+      - run: |
+          oca_export_and_commit_pot
+          git fetch --unshallow --all
+          git push https://x-access-token:{{ "${{" }} secrets.GIT_PUSH_TOKEN {{ "}}" }}@github.com/{{ "${{" }} github.repository {{ "}}" }}
+        if: {{ "${{" }} matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == 'OCA' && startsWith(github.ref, 'refs/heads/{{ odoo_version }}') {{ "}}" }}
```
joao-p-marques, line 65: What do you think of using something like https://github.com/marketplace/actions/github-commit-push to simplify here? The Github Actions from the marketplace are a great way of simplifying and reduce maintenance long term...

Also, the syntax could be a little simplified using [Jinja's `raw` block](https://jinja.palletsprojects.com/en/2.11.x/templates/#escaping). The only Jinja variable used in the if expression is `odoo_version`, but even that variable could be made part of the job variables or, as explained next, even that part of the expression could potentially be removed.

As the job will only run for pull requests and pushes to the main branch, we can simply check if we are in a push event (which means we are necessarily in a push event in the main branch)

```suggestion
      - run: oca_export_and_commit_pot
      - name: Commit & Push changes
        {% raw %}
        if: ${{ matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == 'OCA' }}
        {% endraw %} 
        uses: actions-js/push@master
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

joao-p-marques (commented):
This is a great improvement IMO :+1: 

I just have a small suggestion for the final part of the CI job spec, although it would need more testing...

.github/workflows/test.yml.jinja
sbidoul: Thanks for the tip with the raw block. I'll test it.
Regarding the push action, I think I'll keep it explicit because it's just 2 lines of quite simple code and I don't think its worth pulling a dependency for that, especially a third party one.

simahawk:
FTR this is already live on https://github.com/OCA/odoo-pim/tree/13.0 and https://github.com/OCA/odoo-pim/tree/14.0 and works smoothly :heart:

pedrobaeza:
Is there any solution for runbot thing? Maybe time to go to a newer runbot version with Docker support?

sbidoul:
@gurneyalex is working on the runbot side, the idea is to deploy the latest runbot.

sbidoul:
Ah, and another thing to handle here is the `rebel_module_groups` copier question, to generate addons INCLUDE/EXCLUDE in the job matrix.

Yajo:
Shouldn't this be removing the `.travis.yml` file?

sbidoul:
> Shouldn't this be removing the .travis.yml file?

Not until runbot can work without it, unfortunately... In the meantime we can already test it works correctly, at the expense of some additional carbon emissions.

.github/workflows/test.yml.jinja
```diff
@@ -0,0 +1,65 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - container: {{ IMAGES["odoo"][odoo_version] }}
+            makepot: "true"
+          - container: {{ IMAGES["ocb"][odoo_version] }}
+    services:
+      postgres:
+        image: postgres:9.6
```
Yajo, line 42: I think we should use more modern postgres versions.

At least, I'd use the latest stable for the date of release of each odoo version. You can get the relevant info from https://www.postgresql.org/support/versioning/

.github/workflows/test.yml.jinja
```diff
@@ -0,0 +1,65 @@
...
+  test:
+    runs-on: ubuntu-latest
+    container: {{ "${{" }} matrix.container {{ "}}" }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - container: {{ IMAGES["odoo"][odoo_version] }}
+            makepot: "true"
+          - container: {{ IMAGES["ocb"][odoo_version] }}
```
Yajo, line 39: FWIW you can use macros, which can be in a separate file.

As time goes on, there'll be more little stuff that will be different depending on the base odoo version, so maybe having that info in a separate file could help.

Just a suggestion, it's OK as it is anyway. 😊

.github/workflows/test.yml.jinja
sbidoul: My position about this is that we should test with the lowest python and postgres version that Odoo supports, and rely on python and postgres to assure backward compatibility in more modern versions. But we can certainly discuss this with the wider community.

Yajo: Well, OK. I guess the best would be to just do a matrix with all supported versions, but that has another kind of costs and implications, and in any case it seems outside this PR scope, so we can move forward.

.github/workflows/test.yml.jinja
sbidoul: done

sbidoul:
I added support for rebel module groups. I think this is complete, but before merging I need to deploy a merge bot update to ignore these actions.

.github/workflows/test.yml.jinja
sbidoul: Since this works and is reasonably easy to read I think I'll keep it like this for now. We can refactor later if it becomes too complex.

Yajo:
I don't understand that last comment, but in case you find it difficult to know which actions belong to the template repo and which ones belong to subprojects originated by the template, we can make use of the `_subdirectory` setting and keep all template code separate from the repo code.

It can simplify other things too. You can see [how autopretty does it](https://github.com/copier-org/autopretty/blob/65aeb34041cbe2281c57be0c04635ed2c61c4c80/copier.yml#L1) to understand what I mean.

sbidoul:
> before merging I need to deploy a merge bot update to ignore these actions

Making the merge bot ignore some github checks (such as these tests) and not other (such as pre-commit) is going to make it overly complex, so I'm not going to do that. I'm going to deploy this as is, and try to be available to help resolve the issuses as they arise.

pedrobaeza:
OK, I think we can handle that through the usual general maintainers including me.

Yajo:
Can you please fix the conflicts before reviewing?

Yajo (approved):
It looks very nice.

.github/workflows/tests.yml.jinja
```diff
@@ -0,0 +1,93 @@
...
+        run: manifestoo -d . check-dev-status --default-dev-status=Beta
+      - name: Run tests
+        run: oca_run_tests
+      - uses: codecov/codecov-action@v1
+      {% raw -%}
+      - name: Update .pot files
+        run: |
+          oca_export_and_commit_pot
+          git fetch --unshallow --all
+          git push https://x-access-token:${{ secrets.GIT_PUSH_TOKEN }}@github.com/${{ github.repository }}
```
sbidoul, line 91: Here it would be nice to not fail if we can't do a fast-forward push.

Yajo: You can prepend a `git pull --rebase` before pushing.

sbidoul: I don't think we want to rebase. If a FF push is rejected it means something else has been merged in the meantime and the CI on that other commit is going to push .pot files so we can ignore.

joao-p-marques (approved):

.github/workflows/tests.yml.jinja
```diff
@@ -0,0 +1,93 @@
...
+      - name: Check licenses
+        run: manifestoo -d . check-licenses
+      - name: Check development status
+        run: manifestoo -d . check-dev-status --default-dev-status=Beta
+      - name: Run tests
+        run: oca_run_tests
+      - uses: codecov/codecov-action@v1
+      {% raw -%}
+      - name: Update .pot files
+        run: |
```
Yajo, line 88: Please add here:

```suggestion
        continue-on-error: true
        run: |
```

Docs: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

It should fix @sbidoul's valid concern from https://github.com/OCA/oca-addons-repo-template/pull/61/files#r681616575.

Yajo:
Please rebase.

Also I'm wondering if this is useful until travis can be dumped (follow-up of https://github.com/OCA/oca-addons-repo-template/pull/61#issuecomment-858505324).

sbidoul:
@yajo, do you know how to make this `test.yml` appear only when odoo_version >= 15.0 ?

PR #63, opened 2021-07-22
[ADD] Automatic PR / Issue stale workflow

ivantodorovich:
Following up on these topics:

- https://github.com/OCA/oca-github-bot/issues/10
- https://github.com/OCA/oca-github-bot/pull/67

I re-open the discussion to include this stale workflow in our repositories.

The configuration proposed is just to start the conversation, I expect it to change with whatever comes up from this thread :)

For reference: `actions/stale` docs: https://github.com/actions/stale

@sbidoul I understand from https://github.com/OCA/oca-github-bot/pull/67#issuecomment-605619867 that you've played with it on the mis-builder repository, however I don't see it active for the current default branch, did something go wrong with it or was it just forgotten?

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,74 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+    - cron: "0 12 * * 0"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      # 120+30 day stale policy for PRs
+      # * Except PRs marked as "no stale"
+      - name: Stale PRs policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-pr-labels: "no stale"
+          days-before-stale: 120
+          days-before-close: 30
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          remove-stale-when-updated: true
+          stale-pr-label: "stale"
+          stale-pr-message: >
+            There hasn't been any activity on this pull request in the past 4 months, so
+            it has been marked as stale and it will be closed automatically if no
+            further activity occurs in the next 30 days.
+
+            If you want this PR to never become stale, please ask a PSC member to apply
+            the "no stale" label.
+
+      # 180+30 day stale policy for open issues
+      # * Except Issues marked as "no stale"
+      - name: Stale Issues policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-issue-labels: "no stale,needs more information"
+          days-before-stale: 180
+          days-before-close: 30
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          remove-stale-when-updated: true
+          stale-issue-label: "stale"
+          stale-issue-message: >
+            There hasn't been any activity on this issue in the past 6 months, so it has
+            been marked as stale and it will be closed automatically if no further
+            activity occurs in the next 30 days.
+
+            If you want this issue to never become stale, please ask a PSC member to
+            apply the "no stale" label.
+
+      # 15+30 day stale policy for issues pending more information
+      # * Issues that are pending more information
+      # * Except Issues marked as "no stale"
+      - name: Needs more information stale issues policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          only-labels: "needs more information"
+          exempt-issue-labels: "no stale"
+          days-before-stale: 15
+          days-before-close: 30
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          remove-stale-when-updated: true
+          stale-issue-label: "stale"
+          stale-issue-message: >
+            This issue needs more information and there hasn't been any activity
+            recently, so it has been marked as stale and it will be closed automatically
+            if no further activity occurs in the next 30 days.
+
+            If you think this is a mistake, please ask a PSC member to remove the "needs
+            more information" label.


joao-p-marques (commented):
Hi! I think this is a good idea 👍 

I can’t say much about what would be the better approach for the workflow in the final version, but maybe this could be an optional feature that you can enable with a variable in the copier template? It could even default to `True`, but that way PSCs would have a way of opting out…

Also, it seems like the tests are failing when generating from the jijnja template. If this file won’t have actual jinja templating, you can make it a regular `.yml` file and it will be copied to the destination.

.github/workflows/stale.yml.jinja
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    steps:
+
+      # 120+30 day stale policy for PRs
+      # * Except PRs marked as "don't stale", "help wanted" or "needs review"
+      - name: Stale PRs policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-pr-labels: "don't stale,help wanted,needs review"
```
Yajo, line 18: I'd just use `no-stale` (I hope I'm wrong, but that `'` may be problematic somewhere).

IMHO if "help wanted" but nobody wants to help, it should stale. The same if it "needs review" but nobody reviews. If we have some PRs that need review for v7 and nobody wants them anymore, they should fade away.

Yajo (commented):
I like very much the idea. Thank you.

As @joao-p-marques said, please remove the `.jinja` suffix from the file, as there's no templating here, and github's templating syntax conflicts with copier's.

.github/workflows/stale.yml.jinja
ivantodorovich: I've just changed it to `no stale` :) Spaces are ok as far as I can see, by checking other project's implementation of this action.

As for `help wanted` and `needs review`, yes indeed :) I thought maybe we should start with a less intrusive approach, but in fact we already have `no stale` for this :+1:

ivantodorovich:
> Also, it seems like the tests are failing when generating from the jijnja template. If this file won’t have actual jinja templating, you can make it a regular .yml file and it will be copied to the destination.

> As @joao-p-marques said, please remove the `.jinja` suffix from the file, as there's no templating here, and github's templating syntax conflicts with copier's.

Indeed, thanks!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,47 @@
...
+            further activity occurs in the next 30 days.
+
+      # 60+30 day stale policy for issues pending more information
+      # * Issues that are pending more information
+      # * Except Issues marked as "no stale"
+      - name: Needs more information stale issues policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          only-labels: "needs more information"
```
joao-p-marques, line 36: Not all issues will be marked with this label. For example, old ones that maybe should stale don't have it, so they will remain as they are now... Is that the desired behavior? Maybe just having `no stale` to explicitly avoid it is enough?

joao-p-marques (commented):
Just another small doubt. Otherwise, LGTM :+1:

.github/workflows/stale.yml
ivantodorovich: Indeed, even further this label is not one of the default ones in our repos.. so I will probably never be used.

However I'm not sure what we want here..

Issues are not like PRs, where if the author stops taking care of it we want to close & dismiss. IMO, we only want to close issues if:

- They have been fixed
- They're not really issues (not in the scope of the module, etc)
- They're not clear / unable to reproduce

But if an issue is valid and remains open, no matter for how long, we shouldn't close it I think.

---

On the other hand.. keeping a long list of probably outdated and unqualified issues is not helping newcomers or people eager to contribute to take care of them :sweat_smile:  so maybe it isn't such a bad idea to have them stale unless we qualify them with "no stale"

So you see, I'm not sure what's best here. Opinions?

pedrobaeza: I prefer to close them. Issues for old versions don't add too much value. If the error is still reproducible (and applicable) to newer versions, it can be re-opened retargeting them or create new ones.

ivantodorovich: How about now?

I kept two rules for issues:

1) Applies on all issues: `180+30 day stale`

2) Applies on issues with "needs more information": `60+30 day stale policy`


Reasoning behind rule 2 is that we want to give some time for the author to clearify the issue (so we tag as needs more information), and if he/she doesn't reply in 60 days, we stale it automatically (then close in another 30 days)

joao-p-marques: Ok, that seems like a good compromise :+1: 

About the code, will these 2 rules applying to issues conflict? It looks like they won't, AFAICS, but just to check... You could probably add `needs more information` in the `exempt-issue-labels` key for the `180+30` policy to ensure they won't conflict...

Yajo: About https://github.com/OCA/oca-addons-repo-template/pull/63#discussion_r676637246: I agree with @joao-p-marques and @pedrobaeza. If some issue is valid and yet old, we just have to add the label.

Here in OCA opening issues is not as helpful as in other communities. Usually having an issue is good because the code grows and it should be fixed. However here our code doesn't grow into one direction, but in many (one per version), so as @pedrobaeza said an issue for an older Odoo version can be more a burden than a help.

Also here maintainers only maintain the parts of code they care about. One day a module may stop being used and their issues won't be fixed, so it's better let the module die along with its issues and PRs.

And if some issue is persistent, it's just a matter of replying to the bot once every 3 months. Fair enough.

---

About https://github.com/OCA/oca-addons-repo-template/pull/63#discussion_r676847122, it seems OK to me to stale incomplete issues sooner. Actually I think maybe 60+30 days is too much. I'd say 15+30. I think that, if 2 months later some bot reminds you to give more details about an issue, you just won't remember them. 2 weeks seems better IMHO. OK to that anyway.

.github/workflows/stale.yml.jinja
etobella: Just a question, would it have sense to add the Ready to Merge or the Approved on the exempt labels?

I mean, sometimes, an approved PR is kept there for months until some PSC runs the bot, it is not usual on most of repositories, but it might happen on unusual repos.

.github/workflows/stale.yml
ivantodorovich: I've just pushed https://github.com/OCA/oca-addons-repo-template/pull/63/commits/f6b3ca5f7c933ea923f466dcd978195048dfb92b attending to both @Yajo and @joao-p-marques suggestions :)

ivantodorovich:
btw, it may be a good idea to add the labels we're discussing as default labels for the OCA organization and see if we can propagate them to existing repositories somehow

.github/workflows/stale.yml.jinja
ivantodorovich: Hmmm I think I wouldn't add an exception for those.
Some are labeled as "Approved" or "Ready to merge" while in fact there are unattendeed blocking suggestions. I think we should treat them like any other.

Also having this post the stale message is a way to move things forward if they were forgotten.. that is, if the public interest in the PR remains.

sbidoul:
> btw, it may be a good idea to add the labels we're discussing as default labels for the OCA organization and see if we can propagate them to existing repositories somehow

They are here https://github.com/OCA/maintainer-tools/blob/16636bd84a7577700512f1c6b3c5e8d28f524c2e/tools/set_repo_labels.py#L14-L28

joao-p-marques (approved):

.github/workflows/stale.yml.jinja
Yajo: Yes, I agree with @ivantodorovich.

Besides, our git bot will send some reminders when a PR is in such state, so that use case is already covered. If nobody attends that still, then probably a human should ping another human, or the PR should stale/die just like any other.

Yajo (approved):
I have one more question: what happens if this is defined in several branches? I guess that Github will only obey the scheduled actions defined in the main branch, right? But just to confirm it.

Other than that, I think we can merge.

ivantodorovich:
> I have one more question: what happens if this is defined in several branches? I guess that Github will only obey the scheduled actions defined in the main branch, right? But just to confirm it.

scheduled actions run on the default/main branch afaik

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

> You can schedule a workflow to run at specific UTC times using POSIX cron syntax. **Scheduled workflows run on the latest commit on the default or base branch**. The shortest interval you can run scheduled workflows is once every 5 minutes.

Yajo:
To me this is ready to merge (squashing). @sbidoul I'd like to see your approval here.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,68 @@
...
+          days-before-pr-close: -1
+          remove-stale-when-updated: true
+          stale-issue-label: "stale"
+          stale-issue-message: >
+            There hasn't been any activity on this issue in the past 6 months, so it has
+            been automatically marked as stale and it will be closed automatically if no
+            further activity occurs in the next 30 days.
+
+            If you think this is a mistake and want to keep the issue open, please ask a
+            PSC member to mark it as "no stale".
```
sbidoul, line 48: Why is PSC member intervention mentioned for issues and not PRs ?

ivantodorovich: Added the same message to PRs :+1: and squashed everything

Yajo (approved):

.github/workflows/stale.yml
sbidoul: Ok, thanks. Just to be sure: a simple comment is going to unstale it, correct ?

ivantodorovich: That's right, according to their docs:

> If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart

sbidoul: Ok thanks. So since it was not entirely obvious to me, we could may be replace 

*If you think this is a mistake and want to keep the issue open, please ask a PSC member to mark it as "no stale.* 

by 

*If you want this issue to never become stale, please ask a PSC member to apply the "no stale" label.* ?

ivantodorovich: done :+1:

sbidoul (approved):
Great! Thank you!

sbidoul:
I've made a release. If you want to try it on some repos, you can apply it with `copier -f update`.

PR #64, opened 2021-07-25
Don't prettify .copier-answers.yml

sbidoul:
This makes things more difficult than necessary.


Yajo:
What difficulties do you find?

sbidoul:
Just like other tool-generated files: having pre-commit change them creates unnecessary busywork.

Yajo:
When you update from the template, there are chances that pre-commit configurations have changed, so in any case you are required to run `pre-commit run -a` always.

Then, if you're gonna run it always anyway, what's the extra work that this adds? 🤔

sbidoul:
I know but at the very least it adds cognitive load (e.g. *why does this .copier-answers.yml did change after copier update while I did not change any answer"). Also when doing copier update I like to have one commit for the template update and another for applying pre-commit: that prettifying of .copier-answer.yml creates unnecessary diffs.
We have had this discussion before and I'm still 100% convinced that tool-generated files must be left alone.

Yajo:
I think the correct choice is the opposite.

Maintaining the long list of excludes and having to discuss about it is more burden than having prettified files (which is nice BTW) when, at the end of the day, you have to run `pre-commit run -a` _always anyway_.

But nevermind, this is not so important. 🤷🏼‍♂️

Issue #69, opened 2021-10-01
Prepare Odoo 15


sbidoul:
Upgrade linters, while avoiding unnecessary disruption which make fw/backports difficult.

sbidoul:
@moylop260 et al, can you look at https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pylintrc.jinja and https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pylintrc-mandatory.jinja and suggest new checks to add or move for v15 ?

moylop260:
Hi @sbidoul thanks for you efforts on this

I just got the diff between the pylint-odoo checks enabled in the configuration file vs all the current ones in pylint-odoo

The diff are the following checks:
  - attribute-string-redundant
  - character-not-valid-in-resource-link
  - consider-merging-classes-inherited
  - context-overridden
  - copy-wo-api-one
     - It is not valid for new versions
  - create-user-wo-reset-password
  - dangerous-filter-wo-user
  - dangerous-qweb-replace-wo-priority
  - deprecated-data-xml-node
  - deprecated-openerp-xml-node
  - duplicate-po-message-definition
  - except-pass
  - file-not-used
  - invalid-commit
  - javascript-lint
    - It is supported by pre-commit eslint
  - manifest-maintainers-list
  - missing-import-error
    - It is not needed anymore
  - missing-manifest-dependency
    - It is not updated to new way to use pypi packages name directly
  - missing-newline-extrafiles
  - missing-readme
  - missing-return
  - no-utf8-coding-comment
    - It is not needed anymore
  - odoo-addons-relative-import
  - old-api7-method-defined
  - po-msgstr-variables
  - po-syntax-error
  - renamed-field-parameter
  - resource-not-exist
  - str-format-used
  - test-folder-imported
  - translation-contains-variable
  - translation-positional-used
  - unnecessary-utf8-coding-comment
  - website-manifest-key-not-valid-uri
  - xml-attribute-translatable
  - xml-deprecated-qweb-directive
  - xml-deprecated-tree-attribute

pedrobaeza:
cc @Yajo

Yajo:
> Upgrade linters, while avoiding unnecessary disruption which make fw/backports difficult.

I wonder if we should start using [darker](https://pypi.org/project/darker/) instead of black, to help on this. What do you think?

pedrobaeza:
[...]

PR #70, opened 2021-10-04
[IMP] support Odoo 15.0

Yajo:
General chore to fix https://github.com/OCA/oca-addons-repo-template/issues/69. Some highlights:

- Upgraded all possible pre-commit hooks, without introducing conflicts.
- Upgraded dependencies used by the template itself.
- Fixed pytest.ini to avoid conflicts with tests inside `maintainer-tools`, and make tests work out of the box in vscode.
- Add test to assert that pre-commit hooks install properly on all versions.
- Use python 3.8 for CI workflows in Odoo 15, as that is the version Odoo uses in their runbots.

### DIFF
--- a/src/.travis.yml.jinja
+++ b/src/.travis.yml.jinja
@@ -6,7 +6,11 @@ cache:
 
 python:
   {#- Pin the lowest supported version for each Odoo release #}
+  {%- if odoo_version < 15 %}
   - "3.6"
+  {%- else %}
+  - "3.8"
+  {%- endif %}
 
 addons:
   {#- Pin the lowest supported version for each Odoo release #}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,7 @@ jobs:
         odoo-version:
           - 13.0
           - 14.0
+          - 15.0
     steps:
       # Prepare environment
       - uses: actions/checkout@v2.3.3
@@ -44,7 +45,6 @@ jobs:
             cache ${{ secrets.CACHE_DATE }} ${{ env.PY }} ${{ runner.os }} ${{
             hashFiles('pyproject.toml') }} ${{ hashFiles('poetry.lock') }}
       - run: poetry install
-
       # Let tests issue git commits
       - run: git config --global user.name CI
       - run: git config --global user.email CI@GITHUB
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,11 +4,11 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.1.2
+    rev: v2.4.1
     hooks:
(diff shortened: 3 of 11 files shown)


pedrobaeza:
Are you including the expire bot with latest change from @hbrunn ?

Yajo:
That's not related to odoo 15 and can be pushed in a separate PR. I think that was @hbrunn's intention after all.

pedrobaeza:
Well, at least the expire bot it is, as we are pushing such template to the new branches that will be the default soon.

Yajo:
Done now in #71

joao-p-marques (approved):

src/.pylintrc-mandatory.jinja
```diff
@@ -18,44 +18,76 @@ enable=anomalous-backslash-in-string,
...
+    translation-positional-used,
     translation-required,
+    unnecessary-utf8-coding-comment,
     unreachable,
     use-vim-comment,
+    website-manifest-key-not-valid-uri,
     wrong-tabs-instead-of-spaces,
+    xml-attribute-translatable,
+    xml-deprecated-qweb-directive,
+    xml-deprecated-tree-attribute,
```
sbidoul, line 90: Don't we need to adjust the list of mandatory check per Odoo version ?

src/.travis.yml.jinja
```diff
@@ -6,7 +6,11 @@ cache:
 
 python:
   {#- Pin the lowest supported version for each Odoo release #}
+  {%- if odoo_version < 15 %}
   - "3.6"
+  {%- else %}
+  - "3.8"
```
sbidoul, line 12: Hm. [Not so long ago](https://github.com/odoo/odoo/pull/70398#issuecomment-832601798) Odoo SA's runbot was still -on 3.6. Did that change recently ?

As of now, Odoo 15.0's setup.py still says >= 3.6

The latest "official" [statement](https://github.com/odoo/odoo/issues/76148#issuecomment-916081840) I know of was that it would >= 3.7 but it was not put into action ?

Don't get me wrong, I'd love to be able to use Python 3.8 features and libraries, but on the other hand, publishing OCA addons that are not compatible with some Odoo deployment is not something we should do lightly.

Yajo:
Yes, just check their runbot logs for saas-14.5 and see it's 3.8

src/.travis.yml.jinja
Yajo: We know that odoo says some things and does others. For me, the runbot logs is the most trustable source of truth. Check them out, they're 3.8. I think that we should just test in the same version as Odoo or we'll have different bugs, which is not fun. And after all we have to make a decision and we cannot defer it anymore. 🤷‍♂️

sbidoul:
> @sbidoul do you agree on using Python 3.8?

Yes... until Odoo clarifies it's position, let's use the version that is on runbot.odoo.com.

Issue #77, opened 2021-10-20
Manage .travis.yml and .t2d.yml


sbidoul:
We need to tweak the copier template to remove `.travis.yml` when `ci=='GitHub'` and remove `.t2d.yml` when `ci=='Travis'`.

If a `copier` guru can help with this, that'd be most welcome.

sbidoul:
No that might be an issue with https://github.com/Vauxoo/travis2docker/pull/137

PR #78, opened 2021-10-20
Fix travis.yml and t2d.yml file handling

joao-p-marques:
Fixes https://github.com/OCA/oca-addons-repo-template/issues/77
Work arround https://github.com/copier-org/copier/issues/461

Rename files to the correct jinja syntax
Add task to remove files if the copier update doesn't do it


sbidoul:
@joao-p-marques how do you test that locally ? With copier 5.1, when I run `copier <local template dir>` it still tries to fetch the git master branch.

joao-p-marques:
> @joao-p-marques how do you test that locally ? With copier 5.1, when I run `copier <local template dir>` it still tries to fetch the git master branch.

You can pass a git reference to copier, with `copier -r <GIT_REF>`

Yajo:
FTR if you write a pytest here, you can use this fixture which supports quickly testing even dirty changes you have locally:

https://github.com/OCA/oca-addons-repo-template/blob/5cea348797e800f8b5518001ecb1704fdcb774c4/tests/conftest.py#L29-L53

PR #79, opened 2021-10-21
[FIX] pre-commit action python version

chienandalu:
pre-commit doesn't support python 3.10 yet, so we stall the version in 3.9

cc @Tecnativa

ping @Yajo @joao-p-marques


joao-p-marques:
Fixes https://github.com/OCA/maintainer-quality-tools/issues/700

pedrobaeza:
@sbidoul we need to deploy this patch in all the branches with GH pre-commit. Can you do it?

PR #84, opened 2021-10-26
Use python 3.7 for Odoo 15

sbidoul:
Odoo now officially says python 3.7 is the minimum supported version for Odoo 15. (https://github.com/odoo/odoo/commit/794677fb6a3391379200eb2144a6ed372e89c17a)

So we test on python 3.7 in OCA, to make sure the addons we publish are compatible with Odoo deployments.
If we would not do this, there is a risk that python 3.8+ code creeps into the OCA code base, creating additional support issues.

This also addresses compatibility with travis2docker that does not support python 3.8.

Depends on 
- [ ] setuptools-odoo release https://github.com/acsone/setuptools-odoo/pull/66
- [ ] oca-github-bot update and deployment
- [x] oca-ci images build https://github.com/OCA/oca-ci/pull/14

### DIFF
--- a/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
+++ b/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
@@ -11,7 +11,7 @@ on:
 {%
 set IMAGES = {
   "odoo": {
-    15.0: "ghcr.io/oca/oca-ci/py3.8-odoo15.0:latest",
+    15.0: "ghcr.io/oca/oca-ci/py3.7-odoo15.0:latest",
     14.0: "ghcr.io/oca/oca-ci/py3.6-odoo14.0:latest",
     13.0: "ghcr.io/oca/oca-ci/py3.6-odoo13.0:latest",
     12.0: "ghcr.io/oca/oca-ci/py3.6-odoo12.0:latest",
@@ -19,7 +19,7 @@ set IMAGES = {
     10.0: "ghcr.io/oca/oca-ci/py2.7-odoo10.0:latest",
   },
   "ocb": {
-    15.0: "ghcr.io/oca/oca-ci/py3.8-ocb15.0:latest",
+    15.0: "ghcr.io/oca/oca-ci/py3.7-ocb15.0:latest",
     14.0: "ghcr.io/oca/oca-ci/py3.6-ocb14.0:latest",
     13.0: "ghcr.io/oca/oca-ci/py3.6-ocb13.0:latest",
     12.0: "ghcr.io/oca/oca-ci/py3.6-ocb12.0:latest",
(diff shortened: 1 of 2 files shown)


rvalyi (approved):
same reasoning as @Yajo for having it as a temporary solution.

SplashS:
I've used patch from this PR on https://github.com/OCA/web/pull/2058
Unfortunately, runbot didn't start. What I did wrong?

sbidoul:
I have made an experimental, limited capacity deployment of runboat on 15.0 branches. You should start seeing `runboat/build` statuses on 15.0 PRs.

It's still very rough with some features missing (e.g. the UI is almost existent, users cannot see the build logs, etc). But it works. Capacity is limited to 2 concurrent builds, max 10 started and max 100 in total deployed and ready to start.

Let me know how this goes. In particular if you see runboat builds that are red while the tests are green don't hesitate to ping me.

pedrobaeza:
That's great! Will check and tell you back

pedrobaeza:
I have tested runboat in several 15.0 PRs and works with a reasonable performance. Not sure if it will decrease the same as runbot when there are all the instances up and running. It would be good also to have some kind of "sleep and wake up" strategy for avoiding a full build, but to not reserve resources while not active.

sbidoul:
> It would be good also to have some kind of "sleep and wake up" strategy for avoiding a full build, but to not reserve resources while not active.

This is exactly what it does: the [initialization step](https://github.com/sbidoul/runboat/blob/main/src/runboat/kubefiles/runboat-initialize.sh) clones the repo, installs the dependencies, initializes the database. Then the deployment starts. To conserve CPU and RAM resources, the number of concurrent initialization is limited strongly as this is the heavier step. Then the number of started deployments is limited but with a higher limit, as there are not many connected users at the same time so this use limited CPU resources. Finally we keep a very high number of deployments that are initialized but stopped and ready to start. Since everything is prepared, they start in a few seconds, and I estimate we can have ~10 000 deployments ready to start with a 1TB disk.

If you want to see a little bit more, you can play with the [api](https://runboat.odoo-community.org/docs#/api/controller_status_api_v1_status_get).

pedrobaeza:
That are wonderful news. Congratulations on the great work!

sbidoul:
> That are wonderful news. Congratulations on the great work!

Yep, that is a fun one, although somewhat big as my side projects go :) I'll be looking for contributors ! I'll convert the TODO list of the project into GitHub issues soon.

pedrobaeza:
Great, I think you can transfer it to OCA and we will check to adopt it for internal test instances.

sbidoul:
Closing this as we have a reasonable solution for 15.0 for now. Next step is a larger scale deployment of runboat on OCA infrastructure, which I will work on with @gurneyalex before the end of the year.

Yajo:
Awesome. Thanks!

joao-p-marques:
Thanks again for all the work! From my testing, it is working great :+1:

sbidoul:
I just deployed an update with access to build initialization and runtime logs, and a reactive UI to start/stop builds.

PR #86, opened 2021-10-27
[IMP] add support for new odoo js modules for *.esm.js files

### LINKED ISSUE #85: Problem with eslint in oca/web v15 repository
I have pre-commit failed in PR https://github.com/OCA/web/pull/2058
Link to build
https://github.com/OCA/web/runs/3997888875?check_suite_focus=true

There is problem with ESLint configuration.
Error that I've got:
`  5:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'`

Locally I avoid this error with configuration patch to .eslintrc.yml:
```
parserOptions:
  ecmaVersion: 2017
  sourceType: module

overrides:
  - files:
    - "**/legacy/**/*.js"
    parserOptions:
      sourceType: script
```

Should I include this patch in oca/web PR or it can be made globally for all v15 repositories?

SplashS:
Contributors must place /** @odoo-module **/ formatted js code in files with extension *.esm.js
All other *.js files will be treated as legacy js files
It will allow to support both formats

Fixes #85

### DIFF
--- a/src/.eslintrc.yml
+++ b/src/.eslintrc.yml
@@ -6,6 +6,12 @@ env:
 parserOptions:
   ecmaVersion: 2017
 
+overrides:
+  - files:
+      - "**/*.esm.js"
+    parserOptions:
+      sourceType: module
+
 # Globals available in Odoo that shouldn't produce errorings
 globals:
   _: readonly


pedrobaeza:
Added both sentences in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

Thanks for the enrichment of that guide.

Issue #80, opened 2021-10-21
[13.0] Pre commit fails with "AttributeError: module 'collections' has no attribute 'MutableSet'"


LoisRForgeFlow:
See [this pipeline](https://github.com/OCA/manufacture/runs/3962457025) from https://github.com/OCA/manufacture/pull/665.

Error trace:

```
install pre-commit
/opt/hostedtoolcache/Python/3.10.0/x64/bin/pre-commit run --show-diff-on-failure --color=always --all-files
[INFO] Initializing environment for https://github.com/oca/maintainer-tools.
[INFO] Initializing environment for https://github.com/myint/autoflake.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-prettier.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-prettier:@prettier/plugin-xml@0.7.2,prettier@1.19.1.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-eslint.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-eslint:eslint@6.8.0.
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/asottile/pyupgrade.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-isort.
[INFO] Initializing environment for https://github.com/acsone/setuptools-odoo.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8:flake8-bugbear==19.8.0.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-pylint.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-pylint:pylint-odoo==3.5.0.
[INFO] Installing environment for https://github.com/oca/maintainer-tools.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/myint/autoflake.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/pre-commit/mirrors-prettier.
[INFO] Once installed this environment will be reused.
[...]

PR #89, opened 2021-11-10
Fix pre-commit errors with black

joao-p-marques:
https://github.com/OCA/product-attribute/runs/4165598273?check_suite_focus=true#step:4:112

```
  File "/home/runner/.cache/pre-commit/repowzndsk7_/py_env-python3/bin/black", line 5, in <module>
    from black import patched_main
  File "/home/runner/.cache/pre-commit/repowzndsk7_/py_env-python3/lib/python3.9/site-packages/black/__init__.py", line 52, in <module>
    from typed_ast import ast3, ast27
  File "/home/runner/.cache/pre-commit/repowzndsk7_/py_env-python3/lib/python3.9/site-packages/typed_ast/ast3.py", line 40, in <module>
    from typed_ast import _ast3
ImportError: /home/runner/.cache/pre-commit/repowzndsk7_/py_env-python3/lib/python3.9/site-packages/typed_ast/_ast3.cpython-39-x86_64-linux-gnu.so: undefined symbol: _PyUnicode_DecodeUnicodeEscape
```

Appears to be an error introduced by https://github.com/python/typed_ast with Python 3.9. As explained in https://github.com/python/typed_ast#python-38, this library won't be supporting Python 3.8+

Black fixes this after release 21, but we can't bump the version in older branches without reformatting, so we need to pin the Python version instead.

@Tecnativa
TT32912

ping @Yajo @pedrobaeza @sbidoul

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -13,5 +13,8 @@ jobs:
         with:
           # The pylint-odoo version we use here does not support python 3.10
           # https://github.com/OCA/oca-addons-repo-template/issues/80
-          python-version: "3.9"
+          # We also need to pin to an older version of python for older odoo versions
+          # where we are not using black > 21. Older black versions won't work with
+          # Python 3.9.8+, and we can't bump black without reformatting.
+          python-version: {% if odoo_version < 15 %}"3.9.7"{% else %}"3.10"{% endif %}
       - uses: pre-commit/action@v2.0.0


pedrobaeza (approved):
What a nuisance the things that the maintainers of the libraries do...

We need to unfold this change massively. @sbidoul can you do it or instruct me how to do it in the OCA's server?

sbidoul:
We could also pin python 3.9.7. The problem is due to the GitHub action runners now using python 3.9.8 since last night and the "old" version on black used on branch 14 depends on a library names typed_ast which apparently is not compatible...

sbidoul (approved):

joao-p-marques:
> We could also pin python 3.9.7. The problem is due to the GitHub action runners now using python 3.9.8 since last night and the "old" version on black used on branch 14 depends on a library names typed_ast which apparently is not compatible...

Oh ok, that seems less conflicting, although it is kinda dangerous if `typed_ast` won't get support for this Python version, but again the solution would be to pin to 3.7... :man_shrugging: 

I will change it to `3.9.7` then

joao-p-marques:
> I just wanted to push a test PR into one of the affected repos to check, will do it in the next few minutes...

I couldn't deploy the copier change as is in another repo, but I opened https://github.com/OCA/partner-contact/pull/1176 to test the change manually, and I tried a `copier copy` locally and the file is generated correctly.

src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -13,5 +13,8 @@ jobs:
         with:
           # The pylint-odoo version we use here does not support python 3.10
           # https://github.com/OCA/oca-addons-repo-template/issues/80
-          python-version: "3.9"
+          # We also need to pin to an older version of python for older odoo versions
+          # where we are not using black > 21. Older black versions won't work with
+          # Python 3.9.8, and we can't bump black without reformatting.
+          python-version: "3.9{% if odoo_version < 15 %}.7{% endif %}"
```
pedrobaeza, line 19: Why putting conditions and not directly `3.9.7` for all?

sbidoul: Yes I think we can have 3.9.7 for all < 15, and 3.10 for odoo_version >= 15.

pedrobaeza: I would pin this preventively. We don't need latest Python version for such simple CI. It's not the test one.

joao-p-marques: Ok, I will do as @sbidoul suggests then... But won't https://github.com/OCA/oca-addons-repo-template/issues/80 affect us for v15 branches?

sbidoul: I have tested pylint-odoo 5.0.4 with python 3.10 and it works fine.
An alternative is upgrading black on branches 14 but that will trigger some minor code layout changes, so better not do that now.

joao-p-marques: Ok, I already pushed the change with `3.9.7` for branches bellow v15, and `3.10` for v15.

sbidoul:
Thanks for addressing this! I'll deploy tonight or tomorrow.

sbidoul:
And my motivation for automating https://github.com/OCA/maintainer-tools/pull/486 is increasing :)

sbidoul:
Applied and tested on server-tools 14.0 and 15.0. Deployment in progress on all other repos.

pedrobaeza:
Thanks!

simahawk:
I was about to tackle this and trying to understand why some PRs where suddenly passing :sweat_smile: 
Thanks for the fix! :raised_hands:

tafaRU:
Thanks!

houssine78:
thanks a lot !

PR #90, opened 2021-11-17
[IMP] run oca-gen-addon-readme in pre-commit

ap-wtioit:
This should make it clearer that changes in README.rst are lost if not
 made to the proper readme/*.rst
 
 Info @wt-io-it

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -78,6 +78,8 @@ repos:
       - id: oca-update-pre-commit-excluded-addons
       - id: oca-fix-manifest-website
         args: ["https://github.com/OCA/{{ repo_slug }}"]
+      - id: oca-gen-addon-readme
+        args: ["--repo-name", "{{ repo_slug }}", "--branch", "{{ odoo_version }}", "--no-commit"]
   - repo: https://github.com/myint/autoflake
     rev: {{ repo_rev.autoflake }}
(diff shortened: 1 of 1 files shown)


pedrobaeza:
That was not done, because it requires you to have locally such tool and the dependencies, so it's more startup burden.

ap-wtioit:
@pedrobaeza isn't it already installed with
```
  - repo: https://github.com/oca/maintainer-tools
    rev: dfba427ba03900b69e0a7f2c65890dc48921d36a
```
in .pre-commit-config.yaml?

pedrobaeza:
Oh, it can be... I let the experts in this file to say then.

joao-p-marques (approved):
LGTM from code review :+1:

Yajo (approved):

sbidoul:
There is a problem with this PR.: the maintainer tools version we need to use for this need to be identical to [the one used by the oca-github-bot](https://github.com/OCA/oca-github-bot/blob/daaf1747ca15eab945cec4a97a0919c3c66dd686/Dockerfile#L32), until we find a way to not regenerate all readme files each time the script evolves (https://github.com/OCA/maintainer-tools/issues/520)

So I'm reverting. Can you re-do it with a separate repo entry using the correct maintainer-tools commit ?

Also, note that, similar to #93, the CI check will not error out if the README.rst was not generated by the PR author.

PR #91, opened 2021-11-21
Link to runboat instead of runbot

### LINKED ISSUE #105: Runboat badge instead of runbot one
Similar to #104, but with runboat.

sbidoul:
Closes #105

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "vendor/maintainer-tools"]
-	branch = master
-	path = vendor/maintainer-tools
-	url = https://github.com/OCA/maintainer-tools.git
--- a/src/README.md.jinja
+++ b/src/README.md.jinja
@@ -1,17 +1,7 @@
-{%- macro repo_id(slug) -%}
-    {%- set repos_str -%}
-        {%- include "vendor/maintainer-tools/tools/repos_with_ids.txt" -%}
-    {%- endset -%}
-    {%- for repo in repos_str.split("\n") -%}
-        {%- if repo.endswith("/%s"|format(slug)) -%}
-            {{- repo.split("|")[0] -}}
-        {%- endif -%}
-    {%- endfor -%}
-{%- endmacro -%}
 {%- macro odoo_version_slug(version) -%}
     {{- version | replace(".","-") -}}
 {%- endmacro -%}
-[![Runbot Status](https://runbot.odoo-community.org/runbot/badge/flat/{{ repo_id(repo_slug) }}/{{ odoo_version }}.svg)](https://runbot.odoo-community.org/runbot/repo/github-com-oca-{{ repo_slug }}-{{ repo_id(repo_slug) }})
+[![Runboat](https://img.shields.io/badge/runboat-Try%20me-875A7B.png)](https://runboat.odoo-community.org/builds?repo=OCA/{{ repo_slug }}&target_branch={{ odoo_version }})
 [![Build Status](https://travis-ci.com/OCA/{{ repo_slug }}.svg?branch={{ odoo_version }})](https://travis-ci.com/OCA/{{ repo_slug }})
 [![codecov](https://codecov.io/gh/OCA/{{ repo_slug }}/branch/{{ odoo_version }}/graph/badge.svg)](https://codecov.io/gh/OCA/{{ repo_slug }})
 [![Translation Status](https://translation.odoo-community.org/widgets/{{ repo_slug }}-{{ odoo_version_slug(odoo_version) }}/-/svg-badge.svg)](https://translation.odoo-community.org/engage/{{ repo_slug }}-{{ odoo_version_slug(odoo_version) }}/?utm_source=widget)
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -51,7 +51,7 @@ def test_bootstrap(tmp_path: Path, odoo_version: float, cloned_template: Path):
     # Assert badges in readme; this is testing the repo_id macro
     readme = (tmp_path / "README.md").read_text()
     assert (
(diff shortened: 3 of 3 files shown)


pedrobaeza:
Is runboat to be used for all branches?

sbidoul:
That is a possibility that I'm going to evaluate with @gurneyalex . It supports them all, via the oca-ci images.

For now this is just a draft to show how the link currently [looks like](https://runboat.odoo-community.org/builds?repo=oca/web).

pedrobaeza:
I have removed the draft status by mistake. Ok, the question is due to if not enabled, we should put a conditional version runboat/runbot badge.

Yajo (approved):
I ❤️ it!

Same as @pedrobaeza. If runbot is gonna be killed, then OK. Otherwise, probably better to add the conditional, or to keep both badges until then, marking this one as experimental.

joao-p-marques (approved):
Same as @pedrobaeza and @Yajo 

From my testing, runboat works great as well! Thanks :+1:

sbidoul:
Thanks for the nice feedback folks.

The images work for all branches until 10.0 (see for instance [here](https://runboat.odoo-community.org/webui/builds.html?repo=oca/mis-builder)). And maybe 8 and 9 too (I've not tested but the images exist).

There might be issues with deb and pypi dependencies to fix in older branches but these should be easily fixables by repo maintainers or contributors who need runbot.

So if runboat scales better that runbot (which should be the case), I'm inclined to abandon runbot.

pedrobaeza:
Do you need to have PIP CI for having runboat working?

sbidoul:
It's independent of the CI but it does use pip to install dependencies yes. That simply means that `external_dependencies` must be correct (and, rarely, overrides in setup.py). It also use the `deb` key in `external_dependencies` to install system dependencies. 

That said, runboat does not even know it is deploying Odoo as all knowledge about that is in the [k8s config files and init/start/cleanup scripts](https://github.com/sbidoul/runboat/tree/main/src/runboat/kubefiles). So it should be easily usable with other images than oca/oca-ci.

pedrobaeza:
OK, then let's go aheadt. All aboard in the runboat!!

SplashS:
@sbidoul I have problem with runboat.
There is limit of postgresql connections https://runboat.odoo-community.org/api/v1/builds/bc63cc59d-68a5-49c1-8595-0bcd9838dd1e/init-log

sbidoul:
@SplashS it's a small-scale developement. We are progressively scaling it. I restarted your build.

joao-p-marques:
@sbidoul can we go ahead with this? Is there anything missing?

HaraldPanten:
I have some questions about that, (sorry if I'm missing something or if it's offtopic).

- What would happen with the modules that have been merged? Will them be linked to runboat instead of runboat?
- Are the merged modules going to be "recalculated" with any kind of massive/automatic action in order to be linked to the runboat?

Captura de pantalla 2022-01-11 a las 11 17 29

- What versions does runboat support? I'm pretty sure that it's not going be active for V9, for example.

THX!

sbidoul:
@HaraldPanten the plan is to setup a nginx configuration that redirects the old runbot links to runboat. If anyone wants to help with that it is most welcome.

I also plan to update the addon readme generator to point to runbot, and at the same time setup a mechanism to only regenerate readmes when the readme directory content has changed (https://github.com/OCA/maintainer-tools/issues/520).

sbidoul:
@joao-p-marques to me we can go ahead. Feel free to finish this PR as I don't have much time available these days.

HaraldPanten:
THX @sbidoul

HaraldPanten:
And just to know... Which is the oldest version that runboat will supoort? 🤔

sbidoul:
> And just to know... Which is the oldest version that runboat will support? thinking

It is known to work well back to 10.0. The [oca-ci](https://github.com/oca/oca-ci) image it uses exists for 8 and 9 too, as well as wheels for OCA addons, so it should work for these in theory too. I am myself not too keen to put effort in these old versions, but if anyone wants to help fixing issues that may arise, I'm happy to enable it for some repos on 8 and 9 too.

In practice the main issues I expect is incorrect or missing external_dependencies. All these should be fixable.

All this to say that I'm personally inclined to shut down runbot sooner than later. I'll send a message to the contributors list to ask how people feel about that.

pedrobaeza:
I agree on shutting down runbot.

HaraldPanten:
> > And just to know... Which is the oldest version that runboat will support? thinking
> 
> It is known to work well back to 10.0. The [oca-ci](https://github.com/oca/oca-ci) image it uses [exists](https://github.com/orgs/OCA/packages?repo_name=oca-ci) for 8 and 9 too, as well as wheels for OCA addons, so it should work for these in theory too. I am myself not too keen to put effort in these old versions, but if anyone wants to help fixing issues that may arise, I'm happy to enable it for some repos on 8 and 9 too.
> 
> In practice the main issues I expect are incorrect or missing external_dependencies. All these should be fixable.
> 
> All this to say that I'm personally inclined to shut down runbot sooner than later. I'll send a message to the contributors list to ask how people feel about that.

THX @sbidoul 

It's OK for me if we do not use runboat in old versions. For sure, anybody can cooperate in order to make it possible if it is needed.

+1 for shutting down runbot.

IMO Runboat is much faster for builds and also on it's performance. I like more than runbot and I think it helps functional users to test OCA modules and get more involved to OCA environment.

ValentinVinagre:
nice work :D !!!!

pedrobaeza:
@sbidoul can we merge this?

sbidoul:
> @sbidoul can we merge this?

Now that it is correct (hopefully) and green, yes.

PR #92, opened 2021-11-23
Update black rev in order to include fix for https://github.com/psf/black/issues/1629

tafaRU:
I ran into this [error](https://github.com/tafaRU/l10n-italy/runs/4289069672?check_suite_focus=true#step:4:105) related to https://github.com/psf/black/issues/1629.
In order to fix it, black needs to be updated at least to https://pypi.org/project/black/21.4b0.

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -3,7 +3,7 @@
 
 {%- if odoo_version < 15 %}
   {%- set repo_rev.autoflake = "v1.4" %}
-  {%- set repo_rev.black = "20.8b1" %}
+  {%- set repo_rev.black = "21.4b0" %}
   {%- set repo_rev.eslint = "v7.8.1" %}
   {%- set repo_rev.flake8 = "3.8.3" %}
   {%- set repo_rev.flake8_bugbear = "20.1.4" %}


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 19.10b0
+    rev: 21.4b0
```
simahawk, line 3: this is the repo's file. I guess you want to change this instead https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pre-commit-config.yaml.jinja#L4 ?

tafaRU: Ok, thanks for the advice @simahawk!

sbidoul:
The problem with changing black on active branches is that it creates some churn with PRs becoming red due to minor reformating. That said, black is getting more and more stable over time.

Alternatives:
- using the `# fmt: on/off/skip` black directive to work around local issues
- making the black version a copier question (with same defaults as today), so we can let PSCs manage their black version autonomously across branches

tafaRU:
Ok, thanks for the explanation.
In the meantime I found the lines could make black fail and I fix them. So I can proceed without this.

Personally I'd prefer the following:
> making the black version a copier question (with same defaults as today), so we can let PSCs manage their black version autonomously across branches

PR #95, opened 2021-11-28
Check that pre-commit generated files are in git

### LINKED ISSUE #93: pre-commit check does not detect missing setup.py files
## Describe the bug

When the setup.py files are missing (they are normally generated by the pre-commit config), the pre-commit check running in GitHub actions does not detect the problem.

## To Reproduce

Example: https://github.com/OCA/l10n-italy/pull/2229#issuecomment-979800115

sbidoul:
Fixes #93


sbidoul:
Example run here: https://github.com/OCA/mis-builder/runs/4346459664?check_suite_focus=true

ivantodorovich (approved):
LG

joao-p-marques (approved):

PR #115, opened 2022-02-01
[RFR] Remove except-pass as a required check in pylint-odoo config

StefanRijnhart:
See https://github.com/OCA/pylint-odoo/issues/361


legalsylvain (approved):
Thanks.

pedrobaeza (approved):

sbidoul (approved):

StefanRijnhart:
With @yajo offering the alterative of using `contextlib.suppress` on the pylint-odoo ticket, do we want to proceed with this? I can add a line about this to https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#39exceptions.

Yajo:
Seems a better fix IMHO.

Actually those docs say: *The pass into block except is not a good practice!* but there's no explanation whatsoever, so in any case they can be improved.

legalsylvain:
@StefanRijnhart is it this PR really in draft state ?

PR #116, opened 2022-02-09
[FIX] false failures when detecting unreleased dependencies

Yajo:
The interpreter is launched with `-e` option, which makes bash fail as soon as one command fails.

The grep command will fail when there's no match. That's a valid use case, and means that there are no unreleased dependencies (or no dependencies).

See https://github.com/OCA/server-brand/runs/5123509581?check_suite_focus=true for a false negative. I attach logs here to be able to consult them after github does its GC: [logs_26.zip](https://github.com/OCA/oca-addons-repo-template/files/8031805/logs_26.zip)


Yajo:
Fast-tracking obvious fix.

Issue #103, opened 2021-12-31
Add support for 10, 11, 12


sbidoul:
This is necessary if we want to be able to unplug travis.

TODO

- [ ] extract the flake8 config from MQT
- [ ] extract the pylint config from MQT
- [ ] create a pre-commit config that uses flake8 and pylint with the above configs
- [ ] create a GitHub workflow for tests (the template we have right now should be ok and work for 10, 11, 12 too)

sbidoul:
I think this task is the last one before we can stop Travis and switch all repos to GitHub actions.

If anyone wants to help with it, this is most welcome. I'm happy to provide any guidance that would be necessary.

sbidoul:
@carmenbianca that's great! I have started something in #117. Feel free to pick it up from there. I will not have much time to work on this myself before mid-end march, but let me know if you have any question or doubt on how to move forward.

carmenbianca:
@sbidoul I did a review and some suggestions. I'm a little lost on what specifically is missing/todo. https://github.com/OCA/l10n-italy/pull/2658 demonstrates that #117 mostly works. The test failures seem to be true positives for the maintainers to address.

sbidoul:
> mostly works

"mostly" is the keyword :)

This needs to be made to work and tested in a few representative repos.

PR #117, opened 2022-02-12
[WIP] MQT-compatible pre-commit + GitHub actions for 10, 11, 12

sbidoul:
Towards #103


version-specific/mqt-compat/.flake8
```diff
@@ -0,0 +1,10 @@
+[flake8]
+# E123,E133,E226,E241,E242 are ignored by default by pep8 and flake8
+# F811 is legal in odoo 8 when we implement 2 interfaces for a method
+# F999 pylint support this case with expected tests
+# W503 changed by W504 and OCA prefers allow both
+# E203: whitespace before ':' (black behaviour and not pep8 compliant)
+ignore = E123,E133,E226,E241,E242,F811,F601,W503,W504,E203
+max-line-length = 88
+per-file-ignores=
+    __init__.py:F401
```
carmenbianca, line 10: Does not match https://github.com/OCA/maintainer-quality-tools/blob/master/travis/cfg/travis_run_flake8.cfg#L9

This may be intentional, but flagging it.

version-specific/mqt-compat/.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+  node: "14.13.0"
+repos:
+  - repo: https://github.com/oca/maintainer-tools
+    rev: ab1d7f6
+    hooks:
+      # update the NOT INSTALLABLE ADDONS section above
+      - id: oca-update-pre-commit-excluded-addons
+      - id: oca-fix-manifest-website
+        args: ["https://github.com/OCA/{{ repo_slug }}"]
+  - repo: https://github.com/acsone/setuptools-odoo
```
carmenbianca: A lot of checks are missing here compared to 13.0:

- forbidden-files
- autoflake
- black
- prettier
- pre-commit-hooks
- pyupgrade
- isort

Some of these exclusions seem intentional, but I'm not sure all of them are.

version-specific/mqt-compat/.pylintrc.jinja
```diff
@@ -0,0 +1,100 @@
+[MASTER]
```
carmenbianca, line 1: It'd probably be nice to use the same `{% extends "src/.pylintrc-mandatory.jinja" %}` mechanism here as in `src/`.

carmenbianca (commented):
Mostly LGTM, but some comments.

version-specific/mqt-compat/.pylintrc.jinja
carmenbianca: Proposal in https://github.com/coopiteasy/oca-addons-repo-template/commit/9ef6115740c3b8514d270cc33118de6f1613ab30

version-specific/mqt-compat/.flake8
sbidoul: Yes that can be removed because the old flake8 oca/maintainer-quality-tools is using does not support per-file-ignores.

version-specific/mqt-compat/.pre-commit-config.yaml
sbidoul: That is intentional. My goal is to imitate what oca/mqt does, so we create as little disruption as possible when we are going to apply this to 10,11,12 in order to shut down travis.

version-specific/mqt-compat/.pylintrc.jinja
sbidoul: I've no strong opinion on this. As long as this does the same as the [pylint configs of oca/maintainer-quality-tools](https://github.com/OCA/maintainer-quality-tools/tree/master/travis/cfg).

carmenbianca: It does basically the same, except `.pylintrc` now also repeats the `.pylintrc-mandatory` tests, which I think is desirable for local development.

carmenbianca:
In https://github.com/OCA/product-variant/pull/248 and in local testing flake8 has been, well, flaky. This error keeps popping up:

```
multiprocessing.pool.RemoteTraceback: 
"""
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/flake8/checker.py", line 646, in _run_checks
    return checker.run_checks()
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/flake8/checker.py", line 577, in run_checks
    self.run_ast_checks()
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/flake8/checker.py", line 484, in run_ast_checks
    checker = self.run_check(plugin, tree=ast)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/flake8/checker.py", line 433, in run_check
    return plugin['plugin'](**arguments)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/flake8/plugins/pyflakes.py", line 90, in __init__
    super(FlakesChecker, self).__init__(tree, filename,
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 494, in __init__
    self.handleChildren(tree)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 816, in handleChildren
    self.handleNode(node, tree)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 863, in handleNode
    handler(node)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 1208, in CLASSDEF
    self.handleNode(stmt, node)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 863, in handleNode
    handler(node)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 816, in handleChildren
    self.handleNode(node, tree)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 862, in handleNode
    handler = self.getNodeHandler(node.__class__)
  File "/home/runner/.cache/pre-commit/repo8npub7kc/py_env-python3/lib/python3.9/site-packages/pyflakes/checker.py", line 696, in getNodeHandler
    self._nodeHandlers[node_class] = handler = getattr(self, nodeType)
AttributeError: 'FlakesChecker' object has no attribute 'CONSTANT'
"""
```

pyflakes bug report found here: https://github.com/PyCQA/pyflakes/issues/367

Fixed in pyflakes==2.1.0 (2019-01-23). I'm not sure what the direct relation between pyflakes and flake8 is, but I suspect that any 2019 release of flake8 should not exhibit this problem. (That, or bump down the Python version for the flake8 test.)

Any reason why there's an old version of flake8 (v3.4.1, 2017-07-28) in this PR?

sbidoul:
> Any reason why there's an old version of flake8 (v3.4.1, 2017-07-28) in this PR?

Because that is the one that [MQT uses](https://github.com/OCA/maintainer-quality-tools/blob/0ef63c54d2a75cd25a8283b2c9767db46fe8290e/travis/travis_install_nightly#L6). And if when I tried to use a more recent one, there were new errors because of new checks in recent flake8 versions.

Maybe this can be solved by pinning pyflakes to a version of the same era [here](https://github.com/OCA/mirrors-flake8/blob/v3.4.1/setup.py).

Or by disabling the new checks introduced by flake8 since then. Not sure what's best.

Yajo (approved):
Looks very good.

sbidoul:
@carmenbianca I have squashed my commits here. Feel free to do PRs to my branches, or to cherry-pick my commit in a new PR of yours.

carmenbianca:
@sbidoul I'm afraid I don't have the correct rights to push to your branch. In any case that shouldn't halt process for now :)

carmenbianca:
Notes on the flake8 problem:

1. Lowering Python version in .pre-commit-config.yaml to python3.7 (or lower) → Would work, but would mean that the pre-commit step will fail for everyone who doesn't specifically have python3.7 available on their dev machine.
2. Lowering Python version in .github/workflows/pre-commit.yml to python3.7 (or lower) → Would work, but would mean that the above reported error would still occur locally for anyone who has pre-commit installed against Python ≥3.8.
3. Using flake8 3.4.1 with a more recent pyflakes version (2.1.0) → Untested, and probably won't even install without patching because of [this line](https://github.com/PyCQA/flake8/blob/a4f2135c774c31a7988a7cadbf793fe1c9271284/setup.py#L20) in flake8 3.4.1.
4. Using a later flake8 version and disabling new tests → Would work, but may introduce discrepancies with MQT.

Of these options, 2 and 4 seem the least painful. The reason 2 probably isn't that bad is because I suspect that people doing dev work on Odoo 12 probably do so on Python 3.6/3.7, so they should have the ability to install pre-commit against a lower Python version. But it'll still result in a lot of time wasted à la 'why doesn't this work' for local development.

sbidoul:
In my personal case, 1. would work best. Because I do have several python versions installed on my machine, while my pre-commit install is global and with the latest python version I have.

carmenbianca:
Did a little research on to-disable tests for flake8: the flake8/pyflakes change logs sadly aren't very illuminating. The [pycodestyle change log](https://github.com/PyCQA/pycodestyle/blob/main/CHANGES.txt) is more useful.

On second thought @sbidoul you're probably right that option 1 isn't that bad. Can we reasonably 'force' devs who work on Odoo 12 to have Python 3.6 xor Python 3.7 (depending on which is chosen) installed?

version-specific/mqt-compat/.flake8
```diff
@@ -0,0 +1,10 @@
+[flake8]
+# E123,E133,E226,E241,E242 are ignored by default by pep8 and flake8
+# F811 is legal in odoo 8 when we implement 2 interfaces for a method
+# F999 pylint support this case with expected tests
+# W503 changed by W504 and OCA prefers allow both
+# E203: whitespace before ':' (black behaviour and not pep8 compliant)
+ignore = E123,E133,E226,E241,E242,F811,F601,W503,W504,E203
```
carmenbianca, line 7: F999 doesn't exist. I think F601 is intended?

carmenbianca:
Pushed a partial fix for the failing pre-commit test here: https://github.com/coopiteasy/oca-addons-repo-template/commit/7b4b898fbe40c57c788c7ab7bd65bad6a72b0ddd

However, the solution causes prettier to fail on files like `src/{% if odoo_version > 12 %}.prettierrc.yml{% endif %}` because of https://github.com/prettier/prettier/issues/12143

So I've also pushed https://github.com/coopiteasy/oca-addons-repo-template/commit/087d738011ae2eb39273a81a04deded5116be302 to fix that.

sbidoul:
@carmenbianca will you do another PR with your work on this ?

sbidoul:
This continues in #132

PR #118, opened 2022-02-13
Unconditionally remove .t2d.yml

### LINKED ISSUE #108: Remove .t2d.yml
This .t2d.yml idea did not fly.

Now that we have runboat and we are probably going to shutdown runbot, this must be removed from the template, and we need a task to removed it where it was created.

sbidoul:
closes #108

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -11,9 +11,7 @@ _envops:
   variable_start_string: "{{"
 
 _tasks:
-  # HACK: https://github.com/copier-org/copier/issues/461
-  # prettier-ignore
-  - rm -f {% if ci != "Travis" %}.travis.yml{% endif %} {% if ci != "GitHub" %}.t2d.yml{% endif %}
+  - rm -f .t2d.yml
 
 _migrations:
   - version: v1.3.6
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

copier.yml
```diff
@@ -11,9 +11,7 @@ _envops:
   variable_start_string: "{{"
 
 _tasks:
-  # HACK: https://github.com/copier-org/copier/issues/461
-  # prettier-ignore
-  - rm -f {% if ci != "Travis" %}.travis.yml{% endif %} {% if ci != "GitHub" %}.t2d.yml{% endif %}
+  - rm -f .t2d.yml
```
Yajo, line 14: This won't remove `.travis.yml` if needed. Better:

```suggestion
  - rm -f {% if ci != "Travis" %}.travis.yml{% endif %} .t2d.yml
```

Yajo (changes requested):

copier.yml
sbidoul: Yes, indeed. Thanks!

PR #122, opened 2022-02-22
[IMP] eslint: bump to escmascript 2019

ivantodorovich:
Chosen as [`MAX_ES_VERSION`](https://github.com/odoo/odoo/blob/c16bc3192/odoo/addons/test_lint/tests/test_ecmascript.py#L14) by Odoo since 14.0 [^1]

NOTE: IMO it's not worth adding version-specific templates for this (to keep 2017 version for 13.0), as the list of supported browsers hasn't changed.. it's only that we're all a bit older now

### DIFF
--- a/src/.eslintrc.yml
+++ b/src/.eslintrc.yml
@@ -4,7 +4,7 @@ env:
 
 # See https://github.com/OCA/odoo-community.org/issues/37#issuecomment-470686449
 parserOptions:
-  ecmaVersion: 2017
+  ecmaVersion: 2019
 
 overrides:
   - files:


Yajo (approved):
Seems legit. After all, the fact some Odoo version was released some years ago doesn't mean its uses still use the browsers that got released back then.

flotho (approved):
LGTM

Yajo:
What do you think @Tardo?

legalsylvain (approved):
LGTM. Thanks.

pedrobaeza:
Is this applying to all versions? If so, I think it should be limited to 14.0+

legalsylvain:
Hum...
Well, on one side, @Yajo, you're right when you say "After all, the fact some Odoo version was released some years ago doesn't mean its uses still use the browsers that got released back then.".

In another side, the risk is to make red a some branches (13.0 (and then 10 / 11 / 12, and https://github.com/OCA/oca-addons-repo-template/pull/117 will be merged)). 

if this test fails on the 10.0 web or point-of-sale branch, there is little chance that maintainers will be motivated to correct it.

So maybe better to add a "if v >= 14" test indeed, as @pedrobaeza proposed.

@ivantodorovich : what do you think ?

ivantodorovich:
AFAIK increasing `ecmaVersion` will allow using new JS syntax, but it won't prevent using anything "old", so normally no change should be done in old branches.

As a dev, it'd be a shame having to code in old js syntax, even for 13.0.. when the only limitation is `eslint`, as all browsers supported by Odoo can handle more recent js versions 😅

pedrobaeza:
> As a dev, it'd be a shame having to code in old js syntax, even for 13.0.. when the only limitation is eslint

Are you sure about this assert? I remember the Odoo JS framework limited in the past to use certain features of modern JS.

ivantodorovich:
Odoo Framework might not know how to handle new features, but that doesn't mean you can't use them.

For instance, it's not this case, but when js introduced classes or promises.. old Odoo code won't know how to handle them, so you can't pass them around as arguments to its framework. But you can still use them in your code.

On a general note, new js versions are always backwards compatible.. it would otherwise crash 99% of the internet if it weren't 😅 the only limit here is the browser's support

Tardo:
Enabling it for all versions may be misleading. I agree that it should be for >=13.0. For example, eslint should not allow the use of promises in older versions. It is tempting to mix promises and jquery deferred, but it would result in errors. The good thing for the current change is that this is already happening with 2017.
So, if we don't expect eslint to prevent using unsupported syntax for a particular odoo version I see no reason to limit the change only to Odoo >=13.0.

On the other hand, Odoo <=11.0 uses phantomJS, using something greater than es7 without having it well bounded is dangerous (https://webkit.org/status/).

That said, I don't see any problem in using the latest version of ecmaScript that can be used from Odoo version >=13.0.

simahawk (approved):

Yajo:
El mié, 16 de mar de 2022 a las 15:01:17 PM, Alexandre Díaz 
***@***.***> escribió:
> On the other hand, Odoo <=11.0 uses phantomJS, using something 
> greater than es7 without having it well bounded is dangerous 
> (<https://webkit.org/status/>).
> 

Also, to run tests, you can't predict which chromium version is 
installed alongside newer Odoo installations. Another point to keep in 
mind.

ivantodorovich:
Yes, that's a good point

sbidoul:
Since the template applies to versions >= 13 only, the will only run flake8+pylint for versions < 13 (not eslint etc) I understand the consensus is that this is same enough. So merging. Let me know if I misunderstood some arguments.

Issue #113, opened 2022-01-21
Make manifestoo check not blocking or at least separated


simahawk:
The current manifestoo check it too strict IMO.

You might have a PR which has green tests but that leads to a red build because of this check.
It would be nice to split it a separated check.

A possible use case: you open a PR w/ something experimental on a module that has dev status Beta and you add a dependency on a brand new module as Alpha because you want to test it first in real life together w/ the beta one.

I hope I explained myself correctly :smile: 

We could split the check and make it blocker for PR merge but make the tests run.

pedrobaeza:
I don't get the case. Is it that you forget to put "Alpha" on depending modules? But that's like any other linter: fix it and push it.

simahawk:
mod_a -> Beta, existing since a while
mod_b -> Alpha (because is brand new and you want to keep is alpha until you know it's stable enough)

Create a PR where mod_a it's made dependent on mod_b -> tests fails -> should only block merge but should keep running tests.

pedrobaeza:
Then that's a bug. You can do alpha modules depending on beta ones.

pedrobaeza:
Ah, no, sorry, it's the contrary. beta depends on alpha. That's correct. You should make b Alpha. And no, it shouldn't be that way. You will get a false green, and when trying to merge, finding continuous errors on merging, and you will blame the CI and merge manually...

simahawk:
Why should you blame the CI? You'll see why is red and fix it.

pedrobaeza:
You can do the same on your PR when proposed, not on merge.

sbidoul:
This is a dupe of #83 I think we must make the dev status and license check non blocking on branches < 15 otherwise everything will be red when we deploy GitHub actions and shutdown travis. These issues must be fixed I agree, but it is a new check we are deploying so we need to leave contributors time to fix it and not block everything at once.

We could also make the checks opt-in on branches < 14 and mandatory on >= 15. But I dont know how to do that with copier (i.e; make a question depend on the answer of another question). Maybe @yajo can help here.

sbidoul:
I think I'll move the dev and license checks to separate jobs. Because it does not play well with INCLUDE/EXCLUDE.

sbidoul:
@simahawk not we now have a copier question to disable license and dev status checks.

Issue #83, opened 2021-10-22
[RFC] GitHub actions: make the dev status check optional


sbidoul:
If we deploy GitHub actions on 14 or 13, we should make the development status check non blocking (see for instance https://github.com/OCA/l10n-italy/pull/2468).

This can be done with a copier question.

BT-rmartin:
I would also vote for doing the check on its own job and marking it as optional. Otherwise for old modules, it requires the action on several modules involving their authors/maintainers, as one module could become Stable but the dependencies stay as beta, and right now the check is blocking, when it's something not such critical imho. 
Example: https://github.com/OCA/delivery-carrier/runs/6116809311?check_suite_focus=true

pedrobaeza:
The problem is not that one, as the PR is resolving it. It shouldn't be ignored, as the main goal of development_status will be obfuscated.

sbidoul:
I also think we must make the transition to GitHub actions as friction-less as possible.

So I suggest having a copier question to enable/disable the license and dev status check (so PSCs can choose), and make it mandatory for v15+.

sbidoul:
This was implemented in https://github.com/OCA/oca-addons-repo-template/pull/144

PR #124, opened 2022-02-25
Run dev status and license checks in separate jobs.

sbidoul:
This has the downside of running more jobs, with several parallel
installation of the addons. On the other hand this helps in situations
where INCLUDE/EXCLUDE do not install everything, breaking the
checks with missing dependencies. It should also makes it easier to
surface issues if we make these checks non blocking.

Towards #83 and #113

### DIFF
--- a/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
+++ b/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
@@ -29,6 +29,26 @@ set IMAGES = {
 }
 -%}
 jobs:
+  license:
+    runs-on: ubuntu-latest
+    name: Check licenses
+    container: {{ IMAGES["odoo"][odoo_version] }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install addons and dependencies
+        run: oca_install_addons
+      - name: Check licenses
+        run: manifestoo -d . check-licenses
+  dev-status:
+    runs-on: ubuntu-latest
+    name: Check development status
+    container: {{ IMAGES["odoo"][odoo_version] }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install addons and dependencies
+        run: oca_install_addons
+      - name: Check development status
+        run: manifestoo -d . check-dev-status --default-dev-status=Beta
   unreleased-deps:
     runs-on: ubuntu-latest
     name: Detect unreleased dependencies
@@ -100,10 +120,6 @@ jobs:
           persist-credentials: false
       - name: Install addons and dependencies
         run: oca_install_addons
-      - name: Check licenses
-        run: manifestoo -d . check-licenses
-      - name: Check development status
-        run: manifestoo -d . check-dev-status --default-dev-status=Beta
       - name: Initialize test db
         run: oca_init_test_database
       - name: Run tests


pedrobaeza:
Can this be run only when rebel_modules are declared?

sbidoul:
> Can this be run only when rebel_modules are declared?

This might make things confusing with the checks in different place in different repos (sometimes in separate jobs, sometimes not).

And I think I can improve the install performance, so that should not be an issue in practice.

pedrobaeza:
Well, installing twice things sounds a bit scary, and in practice, INCLUDE/EXCLUDE should be reduced to 0. For me that has been always the lazy way of not dealing with nullifying the problematic part through other ways (like checking in code if in test mode and a special context key for testing just that thing is used). See for example https://github.com/OCA/bank-payment/blob/8.0/account_payment_transfer_reconcile_batch/models/payment_order.py#L29

sbidoul:
Example here: https://github.com/OCA/server-tools/pull/2304

I agree we should try to avoid INCLUDE/EXCLUDE but we do have them, so for now we must support them.

It's not *that* scary :), and I can probably shave one minute off the install step of large repos.

sbidoul:
As that might not be obvious from my OP, let me mention that the installation step I'm talking about is the installation of the dependencies (pip install), not the installation in the database.

pedrobaeza:
Yeah, but some repositories may have large pip requirements, with even compilation. I still think this should be opt-in for those repos with the problem (I don't remember any on v14).

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -29,6 +29,26 @@ set IMAGES = {
...
+      - uses: actions/checkout@v2
+      - name: Install addons and dependencies
+        run: oca_install_addons
+      - name: Check licenses
+        run: manifestoo -d . check-licenses
+  dev-status:
+    runs-on: ubuntu-latest
+    name: Check development status
+    container: {{ IMAGES["odoo"][odoo_version] }}
+    steps:
```
Yajo, line 46: Would it be good to make these optional for some observation time? 🤔 

Just to avoid possible flood of ❌ in CIs.

```suggestion
  license:
    runs-on: ubuntu-latest
    name: Check licenses
    container: {{ IMAGES["odoo"][odoo_version] }}
    continue-on-error: true
    steps:
      - uses: actions/checkout@v2
      - name: Install addons and dependencies
        run: oca_install_addons
      - name: Check licenses
        run: manifestoo -d . check-licenses
  dev-status:
    runs-on: ubuntu-latest
    name: Check development status
    container: {{ IMAGES["odoo"][odoo_version] }}
    continue-on-error: true
    steps:
```

sbidoul: Last time I checked, continue-on-error did show green status, so people will not even pay attention. It's a limitation of GitHub actions that they don't have a warning status.

Yajo: Yes.

Well, that was to avoid too many broken builds, but that's just an hypothesis. We can release an alpha template and check on some selected repos.

yajo:
@sbidoul could you rebase please? to see if this still passes the CI and finally merge? Thanks and sorry for taking so long, I just forgot about this.

legalsylvain:
what is the state of this PR ?

legalsylvain:
@sbidoul : there is conflict to fix. Thanks !

yajo (changes requested):
Rebase needed.

Issue #127, opened 2022-03-04
Pin python version to 3.9 or update dependencies to work with 3.10+?


Tardo:
## Module
.pre-commit-config.yaml.jinja

## Describe the bug
`AttributeError: module 'collections' has no attribute 'MutableSet'`
I have the same issue reported here https://github.com/OCA/oca-addons-repo-template/issues/80 but in my local environment.

## To Reproduce
1. that python3 is pointing to python3.10 on your system 
2. run pre-commit in an OCA repo

**Expected behavior**
pre-commit ends successfully

**Additional context**
Exists a PR to add compatibility to 3.10: https://github.com/OCA/pylint-odoo/issues/360

**Possible temporal fix**
Pin the version to 3.9 instead of 3.x in the `default_language_version` variable

pedrobaeza:
I think pinning the Python version is the better option for now

Yajo:
> Pin the version to 3.9 instead of 3.x in the `default_language_version` variable

If I'm not wrong, pre-commit will not download a 3.9 binary for you. Instead it will fail if it doesn't find a python 3.9 one. That complicates a lot stuff for people on more modern distros.

PR #128, opened 2022-03-07
[FIX] pylint-odoo: usable under python 3.10

Yajo:
Fixes https://github.com/OCA/oca-addons-repo-template/issues/127

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -10,13 +10,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-        with:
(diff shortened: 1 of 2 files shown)


moylop260 (approved):
👍

Tardo (approved):

Tardo:
:+1: Tested. Thanks!

sbidoul:
If we don this, can we / should we remove the [python version pinning here](https://github.com/OCA/oca-addons-repo-template/blob/ee62ad0be6c5c79c3b51bea4328b171fd7e9d75c/src/.github/workflows/pre-commit.yml.jinja#L19-L24) ?

Yajo:
So it seems.

Yajo:
Fixed now.

PR #131, opened 2022-03-16
[IMP] make organization configurable

legalsylvain:
Hi.

First thanks for this great template ! 

I began to use it for grap repo. (and also know that Coop It Easy team do the same). 
I so propose 2 extra simple questions, to make more easy to use this template for other organization. (and also a computed question for the website).

@sbidoul : Do you think it's acceptable ? I set OCA values as default.

@carmenbianca : this should avoid to maintain : 
- https://github.com/coopiteasy/oca-addons-repo-template/pull/1/commits/078f8180ac0a0b6e6d6c1697b7fdaca8c79df834
- https://github.com/coopiteasy/oca-addons-repo-template/pull/1/commits/7abc20425732570e5882f6c1b6e9889354df9c06
- https://github.com/coopiteasy/oca-addons-repo-template/pull/1/commits/9b401a0a2094de6fcfd6e790d41e72b7fcf84640
- https://github.com/coopiteasy/oca-addons-repo-template/pull/1/commits/82b97d7fc32eb023b2102905ee6435e5d67ef836

I applied this template to a grap test branch.
- [here](https://github.com/grap/grap-odoo-incubator/pull/140/commits/b9a551314a11efe2a56a543e3d30df9bc36b2bb9) is the result, that seems OK.
~- I have just a little problem with the change in the test.yml file. generating [that bad lines](https://github.com/grap/grap-odoo-incubator/pull/140/commits/3cfdc4460cf426ba58ebe2cc6d2a126d88404289#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R68). Do you know the correct syntax ?~

thanks for your review.

Note : @sbidoul. If this PR is accepted, (and acceptable), I'll do an update on top of your [PR](https://github.com/OCA/oca-addons-repo-template/pull/117/files).

### DIFF
--- a/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
+++ b/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
@@ -112,5 +112,5 @@ jobs:
       {% raw -%}
       - name: Update .pot files
         run: oca_export_and_push_pot https://x-access-token:${{ secrets.GIT_PUSH_TOKEN }}@github.com/${{ github.repository }}
-        if: ${{ matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == 'OCA' }}
       {%- endraw %}
+        if: {{ "${{" }} matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == '{{ org_slug }}' {{ "}}" }}
--- a/copier.yml
+++ b/copier.yml
@@ -27,6 +27,20 @@ odoo_version:
     - 15.0
   help: Which Odoo version are we deploying in this branch?
 
+org_slug:
+  type: str
+  default: OCA
+  help:
+    What's the organization slug? If you are creating
+    https://github.com/OCA/server-tools, then write "OCA" here.
+
+org_name:
+  type: str
+  default: Odoo Community Association (OCA)
+  help:
+    Tell me the Organization name. It's supposed to be human-readable. It will be used
+    in the author key of the __manifest__ files.
+
 repo_slug:
   type: str
   help:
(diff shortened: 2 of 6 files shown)


sbidoul:
Hi Sylvain. I'm indeed generally prudent to avoid maintenance overhead due to use cases that do not benefit OCA.
But in this case this does not seem to add too much burden, and we already have a configurable organization in several other tools. So :+1: from my side.

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -112,5 +112,5 @@ jobs:
       {% raw -%}
       - name: Update .pot files
         run: oca_export_and_push_pot https://x-access-token:${{ secrets.GIT_PUSH_TOKEN }}@github.com/${{ github.repository }}
-        if: ${{ matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == 'OCA' }}
+        if: ${{ matrix.makepot == 'true' && github.event_name == 'push' && github.repository_owner == {{ org_slug }} }}
```
sbidoul, line 115: Here we are in a `raw` section, so the org slug is not replaced.
The raw section is to avoid escaping the double moustaches of GitHub expression so they are not interpreted by jinja. 
I think I did not have the raw section before you can look in the history. It's gonna be a little bit messy but it's feasible.

legalsylvain: Thanks ! Done.

legalsylvain:
Hi @sbidoul. 

I update my code. This should be all OK.

Note : I run over mis-builder 15.0 branch to see the diff. there is a minor change in the README, due to the rename of OCA to "Odoo Community association (OCA)" but except that point, all is OK.

https://github.com/OCA/mis-builder/pull/425/files

sbidoul (approved):

PR #130, opened 2022-03-14
[FIX] Do not run CI jobs twice on every commit

carmenbianca:
Because of how GitHub workflows works, two jobs were created for every
commit: one for the pull request, and one for the push. This patch
changes that behaviour.

Fixes #129

PR #132, opened 2022-03-17
MQT-compatible pre-commit + GitHub actions for 10, 11, 12

### LINKED ISSUE #103: Add support for 10, 11, 12

carmenbianca:
Duplicate of #117, but from a branch I can write to.

Fixes #103

TODO: Make flake8 behave.


legalsylvain:
Hi @carmenbianca. 
Thanks for your work !

1) Could you rebase on top of head OCA branch. There are 3 missing commits. (not a big deal, but it handles only the removal of  ``.t2d.yml`` file.)
image

2) Don't you have to add keys in the test.yml of the current project ? if yes, you can merge that PR : https://github.com/coopiteasy/oca-addons-repo-template/pull/2

2). I just tested on OCA/pos 12.0 branch. Here is the result : https://github.com/OCA/pos/pull/764
- The result is very red in the pre-commit checks. (https://github.com/OCA/pos/runs/5583735933?check_suite_focus=true)
- the last merge in Travis, 11 days ago, was very green. (https://github.com/OCA/pos/runs/5439320107)

I'm afraid that changing the CI system on these "old" branches requires a lot of work here if you want all the branches to stay green. some choices : 
- keep travis and deploy manually on old branches. (not sure if it is possible, and in any case, it means continue to pay for Travis).
- spend a lot of work on that current branch ``support-10-11-12``.
- assume that CI will be red, after merging that new CI, requiring some works for maintainers. (Optionaly, we could imagine that this year, the money saved on running travis could pay for some maintainer time to make it all green.)

@sbidoul : any point of view regarding that topic as a Board member ?

sbidoul:
> The result is very red in the pre-commit checks. (https://github.com/OCA/pos/runs/5583735933?check_suite_focus=true)

There are 3 kinds of faliure there:

- a technical issue running flake8 (?) to be investigated, but probably fixable and not related to the repo
- some pylint errors, which should not happen if the new pylint config is the same as MQTs ?
- need to fix the website keys in manifest - that can and will be automated if/when we deploy the new pre-commit configs
-

sbidoul:
So in principle, the only kind of errors should be related to invalid external dependencies where the new GitHub actions are stricter.

So I'd suggest digging a little bit more before deciding to keep travis.

legalsylvain:
hi @sbidoul : Thanks for your point of view.

Let's go digging a little bit more. @carmenbianca : do you need some help ? 
@sbidoul : could you merge some trivial / accepted PR to have the possibility to rebase that PR on top of up to date master branch ? (#115, #128, #130, #131 ...)

sbidoul:
@legalsylvain I did some mergin'

Regarding impact on older branches here is a typical example of what needs to be done to fix dependencies: https://github.com/OCA/server-tools/pull/2315. These changes are needed to make runboat work and changes made to please runboat are a prerequisite to please GitHub actions.

That said, runboat being green is not necessary for a merge, so... we may still decide to keep Travis if this proves too difficult.

carmenbianca:
Made some progress.

https://github.com/OCA/pos/pull/773 mostly passes, except some `website` keys were missing/wrong in manifests, and there's a single error being spotted by pylint-mandatory.

I chose to make flake8 run by pinning the Python version to 3.6 in the pre-commit config. This works for the following cases:

- The CI, which I also pinned to 3.6 for version 11 and 12, and version 2.7 for version 10.
- Local development so long as `python3.6` is available.

But it doesn't work for the tests in this repo. I'm not sure which knobs to twist to make that work, but I'll give it a look.

carmenbianca:
Ready for review!

.github/workflows/test.yml
```diff
@@ -19,12 +19,19 @@ jobs:
...
           - 14.0
           - 15.0
+        include:
+          - python-version: 2.7
+            odoo-version: 10.0
+        exclude:
+          - python-version: 3.8
+            odoo-version: 11.0
+          - python-version: 3.8
+            odoo-version: 12.0
```
carmenbianca, line 34: Suggestion: Instead of letting the matrix auto-generate permutations of `python-version` and `odoo-version` and having to deal with incompatibilities, maybe we should hard-code the permutations in `include` to match the permutations in  `src/.github/workflows/test.yml`.

sbidoul: This is the test of the *template*, why does it depend on the Odoo/python version combination ?

carmenbianca:
@sbidoul @legalsylvain this PR needs a review before it becomes stale :)

sbidoul:
Ok, lets merge this and see how it works. Thank you for your work on this @carmenbianca. Much appreciated!

AaronHForgeFlow:
Thank you for the job! :)

Have any of you tested this for version 10.0?

I am stuck in the installation of test-requirements file:

```
+ pip install -r test-requirements.txt
WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
```

There is no test-requirements.txt file in the repo. But if I put one, I get the same error in github actions. Thanks in adavance.

sbidoul:
@AaronHForgeFlow you probably have a mix of Odoo 9 and Odoo 10 addons in your repo ? Please check the versions in the manifests.

AaronHForgeFlow:
@sbidoul  You are right. The modules are installable=False but it seems that is not valid. I will remove those modules. Thank you!

sbidoul:
Hm, if they are installable false they should be ignored.

sbidoul:
That's weird, I'd be interested in more logs about that. Could you open an issue in oca/oca-ci ?

AaronHForgeFlow:
I just checked twice and there was one module with installable=True and version 9.0.1.0.0. So there's no issue with oca/oca-ci . Sorry for the confusion here.

src/.pre-commit-config.yaml.jinja
```diff
@@ -37,7 +37,10 @@
 {%- endif %}
 
 {#- Older versions that differ a lot have their own hardcoded templates for readability #}
-{%- if odoo_version < 14 %}
+{%- if odoo_version < 13 %}
+  {%- include "version-specific/mqt-compat/.pre-commit-config.yaml.jinja" %}
+
+{%- elif odoo_version < 14 %}
```
rousseldenis, line 43: Is it possible that this line breaks updates for version < 13 as there is only one `version-specific/13.0` folder... ?

 @carmenbianca @sbidoul

carmenbianca: Maybe it's a string/float problem? It expects a float where there should be a string, or vice versa? Frankly I'm not sure. I wish I knew how to help.

yajo: You can try replacing `odoo_version` with `odoo_version|float` or `odoo_version|int` in these checks. Just in case you're passing the version as a string for some reason.

Also make sure you're using the latest copier.

PR #136, opened 2022-04-03
Cap python version used for pre-commit

sbidoul:
This is necessary for various reason:
- python 2 for Odoo 10
- python 3.6 for the old flake8 used by branches < 13 in MQT compatibilitly mode
- python < 3.9 for the pylint version used on branch 13.0

This fixes a regression for the 13.0 branches introduced in https://github.com/OCA/oca-addons-repo-template/pull/128 (see example failure in https://github.com/OCA/purchase-workflow/pull/1400).

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -15,6 +15,16 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
+{%- if odoo_version < 11 %}
+        with:
+          python-version: "2.7"
+{%- elif odoo_version < 13 %}
+        with:
+          python-version: "3.6"
+{%- elif odoo_version < 14 %}
+        with:
+          python-version: "3.8"
+{%- endif %}
       - name: Get python version
         run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
       - uses: actions/cache@v1


src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -15,6 +15,16 @@ jobs:
...
+{%- if odoo_version < 11 %}
+        with:
+          python-version: "2.7"
+{%- elif odoo_version < 13 %}
+        with:
+          python-version: "3.6"
+{%- elif odoo_version < 14 %}
+        with:
+          python-version: "3.8"
+{%- endif %}
```
legalsylvain, line 27: that is not exactly the same mapping as here : 
https://github.com/OCA/oca-addons-repo-template/blob/master/src/.github/workflows/%7B%25%20if%20ci%20%3D%3D%20'GitHub'%20%25%7Dtest.yml%7B%25%20endif%20%25%7D.jinja#L15
10 -> 2.7
11 -> 3.5
12, 13, 14 -> 3.6
15 -> 3.8
Is it normal ? 
(naive question. Sorry if it's not relevant.)

legalsylvain (approved):
question inline. Otherwise, LGTM. Thanks !

src/.github/workflows/pre-commit.yml.jinja
sbidoul: Yeah, it's a bit complicated. I'm not sure what's the best approach...
In the end, maybe we should pin the python version in the pre-commit config itself...
But that would be annoying for people who, say, develop with python 3.10 and don't even have older python versions on their system...

carmenbianca: Pinning the Python version in pre-commit makes a lot of sense to me. I think following the mapping referenced by @legalsylvain makes the most sense.

sbidoul: That's fine with me too and probably the most reliable.

Tardo:
I think that version changes made in #128 need be applied manually in https://github.com/OCA/oca-addons-repo-template/blob/master/version-specific/13.0/.pre-commit-config.yaml

sbidoul:
> I think that version changes made in #128 need be applied manually in https://github.com/OCA/oca-addons-repo-template/blob/master/version-specific/13.0/.pre-commit-config.yaml

@tardo, we could but I'm not sure we should. We have seen that the latest pylint version change has caused quite a few red builds on branches 14 and 15. For recent branches, that is fine because people are generally motivated to fix them. But for older branches I think we should try to keep linters as stable as we can. Otherwise they quickly become red and no one is willing to fix them.

Yajo:
But that's probably more a problem related to what linter checks are enabled than related to the python version that executes the linter, right?

Tardo:
You have more knowledge than me.
What is also going to happen is that those who have python 3.10+ will not be able to run pre-commit in their local environments if the dependencies versions are not updated.
I have 3.10 and 3.9 installed right now so I can keep working by setting the python version in the pre-commit configuration and removing the change before the final commit :/

sbidoul:
> But that's probably more a problem related to what linter checks are enabled than related to the python version that executes the linter, right?

Maybe? But I don't quite understand why these `dangerous-view-replace-wo-priority` started showing up.

The more I think of it, the more I think like @carmenbianca that pinning the python version in `.pre-commit-config.yaml` is the more robust and predictable solution. This can be slightly annoying for those who don't have the right python version locally, but at least the error message will be easy to understand (at least easier than random crashes due to incompatibilities). And it is quite easy to install other python versions with pyenv (or the deadsnakes ppa on Ubuntu LTS).

Yajo:
> I don't quite understand why these `dangerous-view-replace-wo-priority` started showing up.

Do you have an example build at hand?

Tardo (approved):
Since it is necessary to unblock this situation, I consider this change to be correct while waiting for a possible "better" solution.

pedrobaeza:
Can you merge and issue a new release for use it on 13.0 branches?

sbidoul:
Will do this weekend

Le ven. 8 avr. 2022 à 20:14, Pedro M. Baeza ***@***.***> a
écrit :

pedrobaeza:
Thanks!

sbidoul:
I deployed yesterday to all 13.0 branches that have a `.copier-answers.yml`.

pedrobaeza:
It seems this is not working at all. I have applied latest template in https://github.com/OCA/commission/commit/857df0f0a8b131184b35d7b10b29f0accb6696da, but it didn't add the `with` sentence, and pre-commit action is red.

I had to add it manually in https://github.com/OCA/commission/commit/1a3a860a725204926d5e55234dcf22968bbccda5, and now pre-commit is fine.

May it be wrong copier syntax or incorrect variable comparison?

PR #139, opened 2022-05-18
Do not ignore lib directory

mt-software-de:
The line "lib/" at the .gitignore file is not really good. 
Because this means no directory with the name lib is added. 
As example see this pr https://github.com/OCA/wms/pull/364
At the module shopfloor_mobile_base of repo https://github.com/OCA/wms
there is the directory static/wms/src/lib and this one would be ignored

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -15,7 +15,6 @@ build/
 develop-eggs/
 dist/
 eggs/
-lib/
 lib64/
 parts/
 sdist/


legalsylvain:
Hi @mt-software-de thanks for your PR.

- removing ``lib`` folder from .gitignore looks great for me. 
- however, adding``.lib`` doesn't makes sense. what is the objective of this one ?

regards

mt-software-de:
I also think it is fine to remove it. But i have thought that maybe you wanted to ignore the root lib/ dir? Should i remove it completly?

src/.gitignore
```diff
@@ -15,7 +15,7 @@ build/
 develop-eggs/
 dist/
 eggs/
-lib/
+.lib/
```
yajo, line 18: I think you can just remove this line.

yajo (changes requested):

mt-software-de:
I removed lib now completely

yajo (approved):

simahawk (approved):

PR #143, opened 2022-06-13
[UPD] pre-commit config for 13.0: bump pylint-odoo version from 3.5.0 to 5.0.5

Tardo:
pre-commit is giving problems with python 3.10 on oca/web due to the version of pylint-odoo used.


pedrobaeza:
This can't be done, as the lines will change, and you're also forcing the use of a higher Python version

Tardo:
So... can fix this https://github.com/OCA/oca-addons-repo-template/blob/master/version-specific/13.0/.pre-commit-config.yaml#L18 to 3.9 ?

Tardo:
Python fixed to 3.9 instead of "latest 3.x"

PR #144, opened 2022-07-12
Adding additional questions to copier template for use outside of the OCA

theangryangel:
The OCA addons repo template is a great resource for all partners, it allows us to build with the possibility of easily offering new modules to the OCA in the future by aligning as close as reasonably possible to the OCA's standards from the outset.

To ease this process we're proposing a few additional questions in the copier template that can disable certain areas that are problematic.

We have set all answers to default to ensure no change in current behaviour. We have setup all (except `github_enable_stale_action`) questions to only be asked when `ci == 'GitHub'`.

* `github_check_license`: Check the manifest license
* `github_enforce_dev_status_compatibility`: Disable the development status, applicable for early work which legitimately may be alpha/pre-alpha
* `github_enable_codecov`: Disable the codecov action
* `github_enable_makepot`: Disable the update of pot files, for not-yet-open-source repositories, this functionality currently fails
* `github_enable_stale_action`: Disable the GitHub stale action

By introducing these questions it allows usage for outside of the OCA without needing to edit github workflows directly.

I've raised this as a draft for discussion, as I'm sure changes will need to be made, tests added, etc. But before we add those I wanted to check if this has the possibility of being accepted, or if it's completely out of scope of the project :)

cc @GlodoUK T3227


legalsylvain:
Divided opinion. I'm un favor to have OCA tools that can be used for other repo. That promote OCA standards and possibly generate new contributors for OCA tools. 
For that reason, i'm OK to integrate makepot questions (useless out of the OCA and maybe in some OCA repo like openupgrade) and codecov that generate extra noice (and there is codecov alternative)
For the two others topics i dont understand why introduce extra questions (and so complexity) 
Why not follow the OCA rules ?

sbidoul:
I'm in favor. These represent a low maintenance cost for OCA and make the template more versatile.

Making the dev status check optional is also important for older OCA branches where making it mandatory would make many repos red, generating unncessary busywork. Could you add a question to enable the license check too for the same reason, and so it closes #83 .

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -101,15 +101,21 @@ jobs:
           persist-credentials: false
       - name: Install addons and dependencies
         run: oca_install_addons
+      {%- if github_check_license %}
```
sbidoul, line 104: The corresponding question is missing ?

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -101,15 +101,21 @@ jobs:
...
       - name: Check licenses
         run: manifestoo -d . check-licenses
+      {%- endif %}
+      {%- if github_dev_status %}
       - name: Check development status
         run: manifestoo -d . check-dev-status --default-dev-status=Beta
+      {%- endif %}
```
sbidoul, line 111: Instead of disabling the check, would you make this step "continue on error" (I'm not sure of the exact syntax) ?

copier.yml
```diff
@@ -127,3 +127,27 @@ include_wkhtmltopdf:
...
     Do you need to install wkhtmltopdf? Usually only needed if you're going to test PDF
     report generation.
+
+github_dev_status:
+  type: bool
+  default: yes
+  help: GitHub action checks the minimum development status?
+  when: &ci_is_github "{{ ci == 'GitHub' }}"
+
+github_codecov:
```
sbidoul, line 137:
```suggestion
github_enable_codecov:
```

copier.yml
```diff
@@ -127,3 +127,27 @@ include_wkhtmltopdf:
   help:
     Do you need to install wkhtmltopdf? Usually only needed if you're going to test PDF
     report generation.
+
+github_dev_status:
```
sbidoul, line 131: nit:

```suggestion
github_enforce_dev_status_compatibility:
```

copier.yml
```diff
@@ -127,3 +127,27 @@ include_wkhtmltopdf:
...
+  help: GitHub action checks the minimum development status?
+  when: &ci_is_github "{{ ci == 'GitHub' }}"
+
+github_codecov:
+  type: bool
+  default: yes
+  help: GitHub action runs codecov/codecov-action?
+  when: *ci_is_github
+
+github_makepot:
```
sbidoul, line 143:
```suggestion
github_enable_makepot:
```

copier.yml
```diff
@@ -127,3 +127,27 @@ include_wkhtmltopdf:
...
+  type: bool
+  default: yes
+  help: GitHub action updates .pot files?
+  when: *ci_is_github
+
+github_stale:
```
sbidoul, line 149:
```suggestion
github_enable_stale_action:
```

legalsylvain:
> Making the dev status check optional is also important for older OCA branches where making it mandatory would make many repos red, generating unncessary busywork. Could you add a question to enable the license check too for the same reason, and so it closes https://github.com/OCA/oca-addons-repo-template/issues/83 .

I didn't thought about old OCA branches. It's a :+1:

sbidoul:
In the same vein, I created #145.

theangryangel:
>  For the two others topics i dont understand why introduce extra questions (and so complexity)

For the stale question, for me, it's 100% selfish, and I'd understand it if that's not something the OCA want to accept.

We don't use GitHub issues for 99% of the repositories until they're open source, and even then the traffic we get on them is currently so low - if we don't need the action running I'd prefer to try and save the planet, even if it's just a very small amount :)

> Could you add a question to enable the license check too for the same reason, and so it closes https://github.com/OCA/oca-addons-repo-template/issues/83 .

Done :)

---

I've adjusted the implementation so that the GitHub stale question doesn't depend on if the ci == 'GitHub', I assumed there are scenarios where Travis is used, but stale does need to be enabled?

I've also tried to work around what I believe is python-poetry/poetry#3010 by slightly adjusting the repo's own test.yml.

Do you want me to add further tests for coverage?

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
theangryangel: I believe I've added the correct incantation for this :)

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
theangryangel: Added

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -101,15 +101,22 @@ jobs:
...
+      {%- if github_check_license %}
       - name: Check licenses
         run: manifestoo -d . check-licenses
+      {%- endif %}
       - name: Check development status
         run: manifestoo -d . check-dev-status --default-dev-status=Beta
       - name: Initialize test db
         run: oca_init_test_database
+        {%- if not github_enforce_dev_status_compatibility %}
+        if: always()
```
sbidoul, line 113: Is it not [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) instead ? 

Can you do the same for the license check ?

And ideally show a test run somewhere ?

theangryangel: > Is it not [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) instead ?

Apologies I thought that was only available at the job level.

> Can you do the same for the license check ?

Added

> And ideally show a test run somewhere ?

I'll set something setup now :)

theangryangel:
Some sample runs are available at https://github.com/GlodoUK/oca_addons-repo-template-pr-144/pulls

copier.yml
```diff
@@ -127,3 +127,32 @@ include_wkhtmltopdf:
   help:
     Do you need to install wkhtmltopdf? Usually only needed if you're going to test PDF
     report generation.
+
+github_enforce_dev_status_compatibility:
+  type: bool
+  default: yes
+  help: GitHub action checks the minimum development status?
+  when: &ci_is_github "{{ ci == 'GitHub' }}"
```
sbidoul, line 135: :heart:

.github/workflows/test.yml
```diff
@@ -54,6 +54,9 @@ jobs:
           key: >-
             cache ${{ secrets.CACHE_DATE }} ${{ env.PY }} ${{ runner.os }} ${{
             hashFiles('pyproject.toml') }} ${{ hashFiles('poetry.lock') }}
+      # HACK python-poetry/poetry#3010 and 3336 workaround for GitHub actions
+      # and poetry 1.1.2 throwing 'Incorrectly nested style tag found.'
+      - run: poetry config experimental.new-installer false
```
sbidoul, line 59: I can't help thinking that poetry is overkill for this project.

theangryangel:
Thanks Stéphane, much appreciated <3

Issue #145, opened 2022-07-13
Add copier question to set CI environment variables


sbidoul:
In some case, it is useful to setup environment variables in the installation/test jobs.

A typical example is the [shopinvader repos](https://github.com/shopinvader/odoo-shopinvader/blob/c0754819fc428fea89ea645f336fe8971436d78d/.github/workflows/test.yml#L31-L32) which need to use a different package index.

sbidoul:
This was done in #179 and #180

PR #146, opened 2022-07-13
Add copier question to set CI environment variables

theangryangel:
Addresses #145, adds the ability to set arbitrary additional environment variables for both Travis and GitHub... I believe the `tojson` filters are the correct solution to quote the env variable keys?

Basing the implementation on:
* https://docs.travis-ci.com/user/environment-variables/#defining-public-variables-in-travisyml
* https://docs.github.com/en/actions/learn-github-actions/environment-variables#using-contexts-to-access-environment-variable-values

cc @GlodoUK T3228

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -156,3 +156,12 @@ github_enable_stale_action:
   type: bool
   default: yes
   help: Create GitHub 'stale' action?
+
+github_ci_extra_env:
+  type: yaml
+  default: {}
+  help: |
+    Any extra environment variables to inject into the CI
+
+    Example: {"KEY": "VALUE"}
+  when: *ci_is_github
--- a/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
+++ b/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
@@ -29,6 +29,7 @@ set IMAGES = {
   }
 }
 -%}
+
 jobs:
   unreleased-deps:
     runs-on: ubuntu-latest
@@ -51,6 +52,12 @@ jobs:
     runs-on: ubuntu-latest
     container: {% raw %}${{ matrix.container }}{% endraw %}
     name: {% raw %}${{ matrix.name }}{% endraw %}
+    {%- if github_ci_extra_env %}
+    env:
+      {%- for key, value in github_ci_extra_env.items() %}
+      {{ key }}: {{ value|tojson }}
+      {%- endfor %}
+    {%- endif %}
     strategy:
       fail-fast: false
       matrix:


src/{% if ci == 'Travis' %}.travis.yml{% else %}.t2d.yml{% endif %}.jinja
```diff
@@ -28,6 +28,13 @@ addons:
       - {{ package|tojson }}
       {%- endfor %}
 
+{%- if ci_extra_env %}
+env:
+  {%- for key, value in ci_extra_env.items() %}
+  - {{ key }}={{ value|tojson }}
+  {%- endfor %}
+{%- endif %}
+
```
sbidoul, line 37: I would not change the Travis template: this fading away and let's not take risk with it.

copier.yml
```diff
@@ -127,3 +127,11 @@ include_wkhtmltopdf:
...
     Do you need to install wkhtmltopdf? Usually only needed if you're going to test PDF
     report generation.
+
+ci_extra_env:
+  type: yaml
+  default: {}
+  help: |
+    Any extra environment variables to inject into the CI
+
+    Example: {"KEY": "VALUE"}
```
sbidoul, line 137: So this would be a github-only question.

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -29,6 +29,14 @@ set IMAGES = {
...
 }
 -%}
+
+{%- if ci_extra_env -%}
+env:
+  {%- for key, value in ci_extra_env.items() %}
+  {{ key }}: {{ value|tojson }}
+  {%- endfor %}
+{%- endif %}
+
```
sbidoul, line 39: Maybe it's better to have this in the `test` job only. Better be specific first and generalize later if needed.

src/{% if ci == 'Travis' %}.travis.yml{% else %}.t2d.yml{% endif %}.jinja
theangryangel: Removed in latest commit

copier.yml
theangryangel: Changed to only ask if ci == 'GitHub' on latest commit

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
theangryangel: Moved to just the test job.

sbidoul (approved):

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -51,6 +52,12 @@ jobs:
     runs-on: ubuntu-latest
     container: {% raw %}${{ matrix.container }}{% endraw %}
     name: {% raw %}${{ matrix.name }}{% endraw %}
+    {%- if github_ci_extra_env %}
+    env:
+      {%- for key, value in github_ci_extra_env.items() %}
+      {{ key }}: {{ value|tojson }}
+      {%- endfor %}
+    {%- endif %}
```
sbidoul, line 60: There is already a `env` key below if there are rebel module groups. So I suspect they will conflict and lead to an invalid yaml?

yajo: The yaml itself should be valid, but I'm not sure about if it's valid configuration for github actions

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -51,6 +52,12 @@ jobs:
     runs-on: ubuntu-latest
     container: {% raw %}${{ matrix.container }}{% endraw %}
     name: {% raw %}${{ matrix.name }}{% endraw %}
+    {%- if github_ci_extra_env %}
+    env:
+      {%- for key, value in github_ci_extra_env.items() %}
+      {{ key }}: {{ value|tojson }}
```
yajo, line 58:
```suggestion
      {{ key }}: {{ value|to_json }}
```

To avoid problems with HTML escaping. https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.tojson

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
sbidoul: @theangryangel do you think you can address this issue and combine the two env sections?

theangryangel: Yup, sorry I let this sit for a bit, I’ll get it sorted first thing tomorrow before I start any other work. Thanks for the prompt

sbidoul:
I'm finishing this one in #179

PR #149, opened 2022-08-22
[IMP] pylint-odoo: upstream pre-commit hook

yajo:
Updating as requested in https://github.com/OCA/pylint-odoo/issues/360#issuecomment-1217528607.

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -15,7 +15,7 @@
   {%- set repo_rev.prettier = "2.1.2" %}
   {%- set repo_rev.prettier_xml = "0.12.0" %}
   {%- set repo_rev.pylint = "v2.11.1" %}
-  {%- set repo_rev.pylint_odoo = "5.0.5" %}
+  {%- set repo_rev.pylint_odoo = "7.0.0" %}
   {%- set repo_rev.pyupgrade = "v2.7.2" %}
   {%- set repo_rev.setuptools_odoo = "2.6.0" %}
 {%- else %}
@@ -31,7 +31,7 @@
   {%- set repo_rev.prettier = "2.4.1" %}
   {%- set repo_rev.prettier_xml = "1.1.0" %}
   {%- set repo_rev.pylint = "v2.11.1" %}
-  {%- set repo_rev.pylint_odoo = "5.0.5" %}
+  {%- set repo_rev.pylint_odoo = "7.0.0" %}
   {%- set repo_rev.pyupgrade = "v2.29.0" %}
   {%- set repo_rev.setuptools_odoo = "3.0.3" %}
 {%- endif %}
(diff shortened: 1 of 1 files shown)


src/.pre-commit-config.yaml.jinja
```diff
@@ -31,7 +31,7 @@
   {%- set repo_rev.prettier = "2.4.1" %}
   {%- set repo_rev.prettier_xml = "1.1.0" %}
   {%- set repo_rev.pylint = "v2.11.1" %}
-  {%- set repo_rev.pylint_odoo = "5.0.5" %}
+  {%- set repo_rev.pylint_odoo = "7.0.0" %}
```
pedrobaeza, line 34: @moylop260 any check has changed in its default behavior during these version?

src/.pre-commit-config.yaml.jinja
```diff
@@ -168,20 +168,16 @@ repos:
       - id: flake8
         name: flake8
         additional_dependencies: ["flake8-bugbear=={{ repo_rev.flake8_bugbear }}"]
-  - repo: https://github.com/PyCQA/pylint
```
pedrobaeza, line 171: No more pylint checks?

yajo: No. For general style we use flake8.

moylop260 (approved):
LGTM 👍

src/.pre-commit-config.yaml.jinja
moylop260: The summary of the [Changelog](https://pypi.org/project/pylint-odoo#CHANGES) is:

```txt
7.0.0
[REF] external-request-timeout: Add ftplib.FTP method (#388)
[FIX] missing-import-error: add xlrd to whitelist (#379)

6.2.0
[IMP] duplicate-po-message-definition: Detecting duplicated only with msgid (#374)
[IMP] test-folder-imported: Consider ‘from . import test’ missing case (#373)

6.1.0
[ADD] external-request-timeout: It could wait for a long time (#370)

6.0.0
[ADD] pre-commit-hooks: Add pylint_odoo hook to be used from pre-commit.yaml directly (#368)
[REF] requirements: Bump pylint package from 2.11.1 to 2.13.5 for py3.x (#367)
[IMP] sql-injection: Detect possible sql injections when using f-strings (#364)
[FIX] Pinned lazy-object-proxy for Python 2.7 (#365)
[FIX] missing-return: Incorrect check for tearDownClass (#362)
```

Notice that f-strings are considered as sql-injection too and other changes for other checks

As @yajo said It shouldn't raise errors but could

If a project is using f-string in a sql-injection way it will be red

Check the other checks changes and check if the configuration files have enabled and so on

But even a sql-injection is better fixed even late than early

pedrobaeza: Yes, I think it's good to improve such check. Thanks for the info.

pedrobaeza:
@sbidoul this is needed for being compatible with Python 3.10, which is the default on Ubuntu 22.04. I think we should deploy it on at least v14 and v15, being that ones the branches under support when 16.0 arises.

yajo:
OK, let's merge this.

Do you guys prefer me to push a beta tag of the template? So you can test on controlled repos before a final release?

pedrobaeza:
As we can update the template selectively on the repo we want, I don't think we need such extra careful on this and I would publish directly a new full version.

yajo:
OK, you got https://github.com/OCA/oca-addons-repo-template/releases/tag/v1.8.0

pedrobaeza:
@sbidoul do you think we can deploy with your script to all 14.0/15.0 OCA branches?

yajo:
I think it's better to update manually a couple of repos and see there's no regression before doing that.

pedrobaeza:
Deployed on https://github.com/OCA/delivery-carrier/pull/526. I will test some other repos.

pedrobaeza:
It went OK also in the most conflicting OCA repository: https://github.com/OCA/sale-workflow/pull/2172, so I think it's safe to deploy this everywhere in 14.0 and 15.0.

PR #151, opened 2022-08-30
Install dependencies with PIP by default

simahawk:
Is not experimental anymore.

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -96,13 +96,12 @@ travis_apt_packages:
   when: *ci_is_travis
 
 dependency_installation_mode:
-  default: OCA
+  default: PIP
   type: str
   choices:
     - OCA
     - PIP
-  help:
-    Choose how to install module dependencies. Right now, "PIP" mode is experimental.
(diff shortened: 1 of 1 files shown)


yajo (approved):
Please check the linter.

sbidoul:
I'm good with this but shouldn't we also change the default ci to GitHub? Travis is kind of deprecated here since we are not going to use it anymore for new repos.

simahawk:
> I'm good with this but shouldn't we also change the default ci to GitHub? Travis is kind of deprecated here since we are not going to use it anymore for new repos.

Yes, my next step :wink:

sbidoul:
What's up with poetry? Any idea,  @yajo ?

sbidoul:
Ah, there was a poetry release a few minutes ago. So you can try capping it to `<1.2` [here](https://github.com/OCA/oca-addons-repo-template/blob/0fc8013f110133d2c33d76133dc6acb0d1a946bb/pyproject.toml#L18).

yajo:
It should require poetry-core instead.

PR #153, opened 2022-09-04
Add missing requirements.txt generator in 13.0

sbidoul:


sbidoul:
cc/ @carmenbianca

pedrobaeza:
I think the reason why this is not activated in 13.0 is because in this version, the external_dependencies manifest key is not using the pip package name, and thus, the requirements.txt will be invalid.

sbidoul:
In 13.0 pypi package names are accepted in external_dependencies. And even in previous versions, the external_dependencies_override must be correct in setup.py otherwise runboat won't work, so it's fine to generate it in 13.0 too. I think I simply forgot to update this file when I interoduced this generator.

pedrobaeza (approved):

carmenbianca (approved):
LGTM

PR #158, opened 2022-09-25
Test with latest version of copier

sbidoul:
We want to test with the latest version of copier.

This project is not meant to be packaged, and we just need to declare test dependencies, so we don't really need a packaging tool for that. So I simplify things a little bit by removing the need for poetry.

A couple of tests need updating for compatibility with the latest version of copier.

Also run tests with python 3.8 and 3.10, and stop testing with 3.6 which is EOL.


yajo (changes requested):
Hi there friend!

This is a surprising move. I think I might be missing some context about *why* replacing poetry.

I don't think this simplifies anything. The workflow looks more complicated to me now, and it doesn't pin any dependencies, so we have no guarantees that devs develop this project with the same versions that other deps or the CI.

If you want to update copier, just run:

```
poetry add copier@latest
```

I don't see any complexity in that.

And if you need a requirements.txt for whatever other reason, then run this:

    poetry export -o requirements.txt

Poetry has the nice ability to be a nice tool for pinning packages, even if you're not developing a package. Without pinning, some day CI will fail and we won't know why.

Besides, @dependabot has builtin support for updating both `pyproject.toml` and `poetry.lock` files. We can drop a configuration file, or enable it in repo settings, and get update PRs automatically.

👎🏼  Personally I'm against this change. OCA should use poetry more, not less IMHO.

sbidoul:
Eh eh, I was expecting that :)

The problem with Poetry, in my view, is that it does not follow python [packaging interoperability standards](https://packaging.python.org/en/latest/specifications/) and has its very how way of doing things. I reckon it was pioneering some solutions when the standards were lagging behind, but now there are better standard based solutions.

Anyway, I don't want to enter a tool war.

Would you mind, then, fixing the tests here, and make them run with the latest version of copier?
I could not do it myself with Poetry (probably because I lack Poetry knowledge).
And we are blocked on that for preparing the 16.0 branches.

sbidoul:
> Besides, https://github.com/dependabot has builtin support for updating both pyproject.toml and poetry.lock files. We can drop a configuration file, or enable it in repo settings, and get update PRs automatically.

BTW, there is no need pinning dependencies here. On the contrary. We want to test with the latest versions all the time, because that is what users are likely to have when they `pipx run copier`.

yajo:
The problem is that this is fixing nothing, but it's introducing new problems.

Keep in mind that you would have a broken Copier 6.2 installation without all of this I'm saying (thanks to dependabot and poetry I detected this incompatibility before the release https://github.com/copier-org/copier/pull/785). Even using pipx. Also, main branch here would be red instead of green if we merged this before. Random CI breakages help nobody.

However, if you prefer another tool that gets the same benefits, I'm not against that.

Let me do the update.

sbidoul:
To test with multiple copier version we should rather add them to the test matrix, no?

yajo:
Yes, but that's not really needed. The template can just support one copier version.

yajo:
Check out https://github.com/OCA/oca-addons-repo-template/pull/159, let's see if that works.

yajo:
> The problem is that this is fixing nothing, but it's introducing new problems.

I want to apologize. Later, while developing #159, I realized this was actually fixing something: adding support for latest copier, and allowing it to coexist with older python versions that couldn't execute copier but had to be installed to execute old hooks. I also realized these words sounded more rude in the comment than in my head. So sorry friend! 😓

I still think the poetry approach is better, due to the reasons already explained. But the rest of the change was actually needed and good, so thanks for your patience, cooperation and dedication. ❤️ 

FWIW I'm using these days [precommix](https://gitlab.com/moduon/precommix), which puts an end to pre-commit surprises. I'd love to see it in use in OCA, but I know it can be scary for some people. It really takes the pre-commit experience to the next level because with it, if pre-commit works today, it will work _forever_.

sbidoul:
No worries, thanks for fixing the CI.

PR #159, opened 2022-09-26
Copier updated

yajo:
- Use newer python versions.
- Add extra older python versions when required by the hooks.
- Fix tests.
- Unpin in `pyproject.toml`, so all you need to update the lockfile is to run `poetry update` from now on.

Closes #158.


yajo:
This is now ready to merge IMHO. Please @sbidoul could you review?

sbidoul:
Thanks. About .gitignore: do you confirm copier supports the symlink? Why not doing different files instead? What we ignore in the template is unrelated to what we ignore in the template itself.

sbidoul (commented):
Just a question about .gitignore.

yajo:
If you prefer we can have separate files, but I have tested that this works as expected in both scenarios.

PR #160, opened 2022-09-26
[IMP] support Odoo 16.0

lmignon:
- Upgrade all possible pre-commit hooks.
- Use python 3.10 for CI workflows in Odoo 16 (same version as used by Odoo in their runbots)
-  Defines docker images to use to run tests with Odoo/OCB 16.0

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -49,7 +49,7 @@
   {%- set repo_rev.pre_commit_prettier = "v3.0.0-alpha.0" %}
   {%- set repo_rev.prettier = "2.1.2" %}
   {%- set repo_rev.prettier_xml = "1.1.0" %}
-  {%- set repo_rev.pylint_odoo = "7.0.0" %}
+  {%- set repo_rev.pylint_odoo = "7.0.1" %}
   {%- set repo_rev.pyupgrade = "v2.38.2" %}
   {%- set repo_rev.setuptools_odoo = "3.1.5" %}
 {%- endif %}
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -12,7 +12,6 @@
   {%- set repo_rev.maintainer_tools = "ab1d7f6" %}
   {%- set repo_rev.nodejs = "14.13.0" %}
   {%- set repo_rev.pre_commit_hooks = "v3.2.0" %}
-  {%- set repo_rev.pre_commit_prettier = "2.1.2" %}
   {%- set repo_rev.prettier = "2.1.2" %}
   {%- set repo_rev.prettier_xml = "0.12.0" %}
   {%- set repo_rev.pylint = "v2.11.1" %}
@@ -29,7 +28,6 @@
   {%- set repo_rev.maintainer_tools = "dfba427ba03900b69e0a7f2c65890dc48921d36a" %}
   {%- set repo_rev.nodejs = "14.18.0" %}
   {%- set repo_rev.pre_commit_hooks = "v4.0.1" %}
-  {%- set repo_rev.pre_commit_prettier = "2.4.1" %}
   {%- set repo_rev.prettier = "2.4.1" %}
   {%- set repo_rev.prettier_xml = "1.1.0" %}
   {%- set repo_rev.pylint = "v2.11.1" %}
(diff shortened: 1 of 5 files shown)


lmignon:
@moylop260, @yajo, ... Do we keep the same pylint-odoo config for 16.0 or do we have rules to make mandatory  and/or new ones to enable

moylop260:
I think we could move all the optional to mandatory for 16.0 but not sure and enable new checks from pylint or pylint-odoo for optional but not sure

src/.pre-commit-config.yaml.jinja
```diff
@@ -28,12 +29,29 @@
...
+  {%- set repo_rev.autoflake = "v1.6.1" %}
+  {%- set repo_rev.black = "22.8.0" %}
+  {%- set repo_rev.eslint = "v8.24.0" %}
+  {%- set repo_rev.flake8 = "3.9.2" %}
+  {%- set repo_rev.flake8_bugbear = "21.9.2" %}
+  {%- set repo_rev.isort = "5.10.1" %}
+  {%- set repo_rev.maintainer_tools = "4cd2b852214dead80822e93e6749b16f2785b2fe" %}
+  {%- set repo_rev.nodejs = "16.17.0" %}
+  {%- set repo_rev.pre_commit_hooks = "v4.3.0" %}
+  {%- set repo_rev.pre_commit_prettier = "v3.0.0-alpha.0" %}
```
yajo, line 49: I would't use an alpha version.

lmignon: @yajo Done

yajo (approved):

lmignon:
@yajo One last question. Do we upgrade the postgresl version into the github action? (currently 9.6)

moylop260:
Good question!

Odoo v16.0 is using json "jsonb_path_query_array"

But it is not available for PSQL <=11.0
  - https://www.postgresql.org/docs/11/functions-json.html

But PSQL>= 12.0
- https://www.postgresql.org/docs/12/functions-json.html

or the same than runbot

moylop260:
Running the following code as server action in runblt.odoo.com branch 16.0

```python
env.cr.execute ("SELECT VERSION();")
res = env.cr.fetchone ()
raise UserError(res)
```

the result is:
  - `PostgreSQL 12.12`

But current stable version is 14

odoo-sh is using PSQL-10 for Odoo <=15.0

Odoo==16.0 is not available in odoo-sh yet in order to check what PSQL version it will use since PSQL-10 will not be compatible anymore

lmignon:
I propose to switch to pg12 to be aligned with odoo.

yajo:
This PR is about supporting v16 in the repo template. For CI workloads you should open the issue in https://github.com/OCA/oca-ci if I'm not wrong.

lmignon:
@yajo a little change was also required into the github workflow to use the right image of postgresql. https://github.com/OCA/oca-addons-repo-template/pull/161

sbidoul:
So what do we do for the pylint-odoo checks?

pedrobaeza:
What are the new checks?

IMO this issue should be solved: https://github.com/OCA/pylint-odoo/issues/387, and the version we put on the scaffolding includes it.

sbidoul:
@moylop260 gentle nudge. What are the new pylint-odoo checks we can enable in `.pylintrc` ?

moylop260:
I'm not sure what is "new" from the OCA perspective

I mean, there are a lot of helpful lints disabled in the OCA

In my case I do the following:
1. Get all the current checks: `pylint --list-msgs`
2. Enable all them and run over all the OCA and custom modules and check what is good and what is not good `pylint -e all -d NOGOOD...`

You can see the enable/disable checks from our configuration files:
 - https://github.com/Vauxoo/pre-commit-vauxoo/blob/main/src/pre_commit_vauxoo/cfg/.pylintrc
 - https://github.com/Vauxoo/pre-commit-vauxoo/blob/main/src/pre_commit_vauxoo/cfg/.pylintrc-optional

sbidoul:
Hi @moylop260 my question was which new *pylint-odoo* checks that have been added recently to and are not listed in our current configs.

Anyway, if no-one is proposing anything concrete, I'll create the 16.0 branches with the same configs as 15.

moylop260:
Hi @sbidoul 

I got it

Could you send me the current pylint-odoo checks `enabled`, please?

In order to be used with the following sentence:
 - `current = {msg[1] for msg in pylint_odoo.get_all_messages().values()}`

So, I will be able to run
`enabled - current`

sbidoul:
@moylop260 here is the current list: https://github.com/OCA/mis-builder/blob/f2c757fc25755129ffca56f37278cafddcc759ee/.pylintrc#L25-L116

sbidoul:
Thanks! I created https://github.com/OCA/oca-addons-repo-template/pull/163

Yes it's great to add new checks to pylint itself when they are not odoo-specific.

moylop260:
I meant,

Enable new checks in the OCA cfg files that were added to pylint core (not only new checks added to pylint-odoo)

because of pylint core has new good ones but OCA didn't analyze what could be helpful

I could create a new issue related to this with new checks identified

moylop260:
FYI https://github.com/OCA/oca-addons-repo-template/issues/164

PR #161, opened 2022-09-27
[IMP] support Odoo 16.0

lmignon:
Odoo v16.0 is using jsonb_path_query_array function available from postgresql >= 12.0


lmignon:
ping @moylop260

pedrobaeza (approved):

moylop260 (approved):
Thank you!

It could be great if we can use the same sub-version than Runbot 12.12

but if it is not possible from gh-actions then 👍

Issue #176, opened 2022-12-08
Add pre-commit hook to check that `*/i18n/en.po` are absent


sbidoul:
The presence of `en.po` files can cause issues. See for example: https://github.com/OCA/knowledge/pull/381

This could be easily implemented with a [fail](https://pre-commit.com/#fail) hook.

Alternatively, or complementarily for 17+, this can be added to https://github.com/OCA/odoo-pre-commit-hooks.

PR #177, opened 2022-12-10
[IMP] add fail for en.po files

etobella:
Adds changes for #176

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,24 +9,31 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{matrix.machine}}
     strategy:
       matrix:
         include:
           - odoo-version: 10.0
             python-version: 2.7
+            machine: ubuntu-20.04
           - odoo-version: 11.0
             python-version: 3.6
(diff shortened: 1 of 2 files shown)


etobella:
@sbidoul setting python 3.6 failedgithub.com:tegin/oca-addons-repo-template.git due to the change of the default machine of ubuntu on github actions, as you can see here:
https://github.com/actions/setup-python/issues/544

Last commit changes fixed it. Also, if we reexecute master now, it will fail without the last commit changes :cry: 
https://github.com/OCA/oca-addons-repo-template/actions/runs/3468866261/jobs/6204635946

sbidoul:
Let's pin ubuntu-20.04 for 10.0 too. And 22.04 for 14-16?

etobella:
@sbidoul Done!

sbidoul (approved):

sbidoul:
Thanks @etobella. I'm launching a mass update of branches 14, 15, 16 and removing en.po files at the same time.

Would you mind doing a similar PR in https://github.com/OCA/oca-addons-repo-template/tree/v1.12.0/version-specific for branche 10 - 13 ?

PR #178, opened 2022-12-11
[FIX] add en.po error for 13.0

etobella:
I noticed, I only did the change for 14, 15 and 16. Applied the same for 13 :disappointed:


sbidoul:
Thanks!

PR #181, opened 2022-12-14
[FIX] use ubuntu 20.04 for odoo<14 ci

hbrunn:
this should fix builds for old Odoo versions since `ubuntu-latest` was switched to 22.04

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -11,7 +11,11 @@ on:
 
 jobs:
   pre-commit:
+{%- if odoo_version < 14 %}
+    runs-on: ubuntu-20.04
+{%- else %}
     runs-on: ubuntu-latest
+{%- endif %}
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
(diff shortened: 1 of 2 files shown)


pedrobaeza:
ping @sbidoul @carmenbianca

src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -11,7 +11,11 @@ on:
 
 jobs:
   pre-commit:
+{%- if odoo_version < 14 %}
+    runs-on: ubuntu-20.04
+{%- else %}
     runs-on: ubuntu-latest
```
sbidoul, line 17:
```suggestion
    runs-on: ubuntu-22.04
```

Let's pin now so we don't have to do it in 2 years?

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -50,7 +50,11 @@ jobs:
               fi
           done
   test:
+{%- if odoo_version < 14 %}
+    runs-on: ubuntu-20.04
+{%- else %}
     runs-on: ubuntu-latest
```
sbidoul, line 56:
```suggestion
    runs-on: ubuntu-22.04
```

PR #187, opened 2023-02-16
[UPD] make the template compatible with the latest version of copier

Tardo:
Changes required for proper operation with the latest version of copier 7.0.1

cc @yajo

### DIFF
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 
 import yaml
-from copier import copy
+from copier.main import run_auto
 
 # Change this one in case it becomes a mandatory check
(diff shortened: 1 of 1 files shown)


yajo (approved):
I think you should [specify `_min_copier_version` in the template](https://copier.readthedocs.io/en/stable/configuring/#min_copier_version) too.

yajo:
BTW, is it really necessary to add all those `|int` things? The choice is already a float, so it should allow comparing without problem out of the box IIUC: https://github.com/OCA/oca-addons-repo-template/blob/3ccb33664717dbd7fb52c10cc44b7545d4e2827c/copier.yml#L21-L32

PR #188, opened 2023-03-01
[FIX] pre-commit: Enforce last version of pre-commit for 14

etobella:
Today, pre-commit on 14.0 branches started to fail:

https://github.com/OCA/account-financial-tools/actions/runs/4301519746/jobs/7498840452


yajo:
Thanks! I'll do a new release now.

PR #189, opened 2023-03-23
Test only supported Odoo versions

SirTakobi:
Implementing what suggested in
> The best bet is to modify the OCA CI to only use OCB. At the end, there won't be any counterpart, as there are no more Odoo updates, so if right now is green, the only chance to break it is on OCB. And we will gain less energy consumption in the builds.

_Originally posted by @pedrobaeza in https://github.com/odoo/odoo/issues/115782#issuecomment-1474877996_

The issue fixed in the linked PR is blocking many good PRs in `12.0` out there.

Here I allow to choose if tests should be executed in Odoo/OCB.

I don't know how to test this locally so I just wrote what seemed appropriate, any feedback is appreciated.
I tried removing the old `.copier-answers` from a repository and `copier /path/to/oca-addons-repo-template .` but it isn't using the edited template that I have locally.

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -76,6 +76,15 @@ ci:
     - GitHub
   help: Which CI system to use ?
 
+odoo_test_flavor:
+  type: str
+  default: Both
+  choices:
+    - Odoo
+    - OCB
+    - Both
+  help: Which Odoo flavor should be used for CI tests ?
+
 travis_apt_sources:
   type: yaml
   default: []
--- a/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
+++ b/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
@@ -63,29 +63,41 @@ jobs:
         {%- if rebel_module_groups %}
         include:
         {%- for group in rebel_module_groups %}
+            {%- if odoo_test_flavor in ["Odoo", "Both"] %}
           - container: {{ IMAGES["odoo"][odoo_version] }}
             include: "{{ group }}"
-            makepot: "{{ github_enable_makepot | bool | lower }}"
             name: test with Odoo
+            {%- endif %}
+            {%- if odoo_test_flavor in ["OCB", "Both"] %}
           - container: {{ IMAGES["ocb"][odoo_version] }}
             include: "{{ group }}"
             name: test with OCB
+            {%- endif %}
+            makepot: "{{ github_enable_makepot | bool | lower }}"
         {%- endfor %}
+            {%- if odoo_test_flavor in ["Odoo", "Both"] %}
           - container: {{ IMAGES["odoo"][odoo_version] }}
             exclude: "{{ rebel_module_groups|join(',') }}"
-            makepot: "{{ github_enable_makepot | bool | lower }}"
             name: test with Odoo
+            {%- endif %}
+            {%- if odoo_test_flavor in ["OCB", "Both"] %}
           - container: {{ IMAGES["ocb"][odoo_version] }}
             exclude: "{{ rebel_module_groups|join(',') }}"
             name: test with OCB
+            {%- endif %}
+            makepot: "{{ github_enable_makepot | bool | lower }}"
         {%- else %}
         {#- If all modules can get along, we test and generate .pot all at once #}
         include:
+            {%- if odoo_test_flavor in ["Odoo", "Both"] %}
           - container: {{ IMAGES["odoo"][odoo_version] }}
-            makepot: "{{ github_enable_makepot | bool | lower }}"
             name: test with Odoo
+            {%- endif %}
+            {%- if odoo_test_flavor in ["OCB", "Both"] %}
           - container: {{ IMAGES["ocb"][odoo_version] }}
             name: test with OCB
+            {%- endif %}
+            makepot: "{{ github_enable_makepot | bool | lower }}"
         {%- endif %}
     services:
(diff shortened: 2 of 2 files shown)


legalsylvain (changes requested):
The idea look great :+1: ! I'm always OK to reduce the impact of the CI when tests are unnecessary.

About the selection of the version, 13.0 is not a good choice for the time being (and maybe 12.0)
See : 
- last V10 commits : https://github.com/odoo/odoo/commits/10.0 : May 18, 2021. (~2 years ago, :ok:)
- last V11 commits : https://github.com/odoo/odoo/commits/11.0  : Jan 20, 2022. (14 monthes ago, :ok:)
- last V12 commits : https://github.com/odoo/odoo/commits/12.0  : Oct 13, 2022 (6 monthes ago, :+1: / :-1: )
-  last V13 commits : https://github.com/odoo/odoo/commits/13.0 : Mar 13, 2023 (~10 days ago, totally :no_entry_sign: )


as a safe choice, I would therefore propose the following rule: 
We can remove the odoo "X.0" branch from the OCA CI, and use only OCB if there is no commit on the "X.0" branch for one year.

So for the time being : 10.0 and 11.0. (and remove 12.0 in October 2023, if no new commits are pushed).

What do you think ?

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -63,27 +61,33 @@ jobs:
         {%- if rebel_module_groups %}
         include:
         {%- for group in rebel_module_groups %}
+            {%- if odoo_version in IMAGES["odoo"] %}
```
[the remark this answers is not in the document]
SirTakobi: Agreed, didn't see that.
I actually don't know if it will work, but I just moved those to the OCB item.

legalsylvain (approved):
for me makes sense. But I'm not expert of that repo, so other review are welcome.

thanks for the changes !

yajo (approved):
LGTM. @sbidoul can you please confirm that the makepot change should behave as expected?

sbidoul:
I think it would be more flexible to add a copier question to enable odoo and/or ocb tests.
This way the odoo test can be disabled by default on older branches but people can still enable them if they want to.

SirTakobi:
> I think it would be more flexible to add a copier question to enable odoo and/or ocb tests. This way the odoo test can be disabled by default on older branches but people can still enable them if they want to.

Good idea, that would be more flexible for sure, I should look up how to do that.
But other than possible technical problems, how should the question behave?

- Choose Odoo/OCB for **all** versions?
  Maybe is too extreme
- Or choose Odoo/OCB for **each** version?
  Dynamic number of questions: is that even possible?
- Or something in between?
  Maybe we ask only for specific versions, like the ones I am already excluding here

sbidoul:
I think we can have a question such as "Which Odoo flavor to test?" with 3 possible answers: odoo/OCB/both.

bealdav:
Good initiative here.

SirTakobi:
:thinking: sorry for the multiple review requests, apparently I can't re-request review from both @yajo and @legalsylvain.

Anyway, I re-requested the reviews because something changed: now the user can choose what to enable, let me know what you think :smile:
cc @sbidoul

sbidoul:
@sirtakobi does it enable makepot in case OCB only is selected?

SirTakobi:
> @SirTakobi does it enable makepot in case OCB only is selected?

Yes because `makepot` is added to the last configuration, regardless of what was selected

sbidoul:
@SirTakobi we must ensure to not run makepot twice. So it should appear in OCB jobs only it the flavor is OCB.

SirTakobi:
> @SirTakobi we must ensure to not run makepot twice. So it should appear in OCB jobs only it the flavor is OCB.

Looking at the code changes, I think it will only be run once; of the three possible choices:

- "Odoo": `makepot` is run in "Odoo"
- "OCB": `makepot` is run in "OCB"
- "Both": `makepot` is run in "OCB"

Have you noticed some case in the code that I'm missing?

sbidoul:
Ah, yes, sorry I misunderstood. :+1:

sbidoul:
Ah, yes, sorry I misunderstood. :+1:

sbidoul (approved):

yajo:
Thanks!

Would you please apply this into one or two repos before we do a release? So we can see it works in the real world. Update to latest unreleased change with:

    copier -r HEAD update

SirTakobi:
> Thanks!
> 
> Would you please apply this into one or two repos before we do a release? So we can see it works in the real world. Update to latest unreleased change with:
> 
> ```
> copier -r HEAD update
> ```

Done in https://github.com/OCA/l10n-italy/pull/3241

copier.yml
```diff
@@ -76,6 +76,15 @@ ci:
     - GitHub
   help: Which CI system to use ?
 
+odoo_test_flavor:
+  type: str
+  default: Both
```
sbidoul, line 81: Do we want to set the default value to OCB for Odoo < 13 ?

SirTakobi:
Yes that might be useful

copier.yml
sbidoul: We already have `default: "https://github.com/{{ org_slug }}/{{ repo_slug }}"` so we probably can put any python expression in there?

yajo: Yes, you can do that.

PR #190, opened 2023-03-28
feat: generate addons readme in pre-commit

yajo:
Devs sometimes just update the `readme/*.rst` portions, relying in our nice bot that will convert that to a real `README.rst` + `index.html` after merge.

When functional people go test PRs in runboat, the 1st thing they look at is the README. That's OK because there's where the docs should explain what the module is doing.

These 2 situations are good in their own terms, but when you merge both, you get to a problem. You have to either:

- Teach functional people to review the PR code, go to the `readme/` folder and understand the `.rst` syntax and the advanced github diff tooling.
- Teach devs to remember always using the `oca-gen-addon-readme` before pushing the PR, so functionals can review the README in a more human-friendly fashion.

None of these is a good solution. Both add manual and complex steps to the contribution workflow.

Quoting from [OCA blog][1]:

> The Board are working on these Major Projects for 2023:
> Improve Contributor onboarding journey, leverage delegates
> - Including Functional Consultants

This PR goes along with those goals. By moving the readme generation into a pre-commit hook:

1. The dev just needs to stage -> commit -> re-stage -> re-commit. Just like always.
2. The functionals have a nicely rendered README available right from the Odoo UI. That's the UI a functional needs to know deeply (not Github's).

@moduon

[1]: https://odoo-community.org/blog/news-updates-1/oca-newsletter-1-2023-136

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -109,6 +109,11 @@ repos:
     hooks:
       - id: oca-checks-odoo-module
       - id: oca-checks-po
+      - id: oca-gen-addon-readme
+        args:
+          - --addons-dir=.
+          - --branch={{ "%.01f" | format(odoo_version) }}
+          - --repo-name={{ repo_slug }}
   - repo: https://github.com/myint/autoflake
     rev: {{ repo_rev.autoflake }}
     hooks:
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -43,7 +43,7 @@
   {%- set repo_rev.flake8 = "3.9.2" %}
   {%- set repo_rev.flake8_bugbear = "21.9.2" %}
   {%- set repo_rev.isort = "5.12.0" %}
-  {%- set repo_rev.maintainer_tools = "4cd2b852214dead80822e93e6749b16f2785b2fe" %}
+  {%- set repo_rev.maintainer_tools = "58557dd6b2c72e698dd2a0b4c5b435815e5d9d02" %}
   {%- set repo_rev.nodejs = "16.17.0" %}
   {%- set repo_rev.odoo_pre_commit_hooks = "v0.0.25" %}
   {%- set repo_rev.pre_commit_hooks = "v4.3.0" %}
@@ -104,16 +104,17 @@ repos:
       - id: oca-update-pre-commit-excluded-addons
       - id: oca-fix-manifest-website
         args: ["{{ repo_website }}"]
-  - repo: https://github.com/OCA/odoo-pre-commit-hooks
-    rev: {{ repo_rev.odoo_pre_commit_hooks }}
-    hooks:
-      - id: oca-checks-odoo-module
-      - id: oca-checks-po
       - id: oca-gen-addon-readme
         args:
           - --addons-dir=.
           - --branch={{ "%.01f" | format(odoo_version) }}
           - --repo-name={{ repo_slug }}
+          - --if-source-changed
+  - repo: https://github.com/OCA/odoo-pre-commit-hooks
+    rev: {{ repo_rev.odoo_pre_commit_hooks }}
+    hooks:
+      - id: oca-checks-odoo-module
+      - id: oca-checks-po
   - repo: https://github.com/myint/autoflake
     rev: {{ repo_rev.autoflake }}
     hooks:
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -43,7 +43,7 @@
   {%- set repo_rev.flake8 = "3.9.2" %}
   {%- set repo_rev.flake8_bugbear = "21.9.2" %}
   {%- set repo_rev.isort = "5.12.0" %}
-  {%- set repo_rev.maintainer_tools = "58557dd6b2c72e698dd2a0b4c5b435815e5d9d02" %}
+  {%- set repo_rev.maintainer_tools = "969238e47c07d0c40573acff81d170f63245d738" %}
   {%- set repo_rev.nodejs = "16.17.0" %}
   {%- set repo_rev.odoo_pre_commit_hooks = "v0.0.25" %}
   {%- set repo_rev.pre_commit_hooks = "v4.3.0" %}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,3 +14,12 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v2
       - uses: pre-commit/action@v2.0.0
+        env:
+          # Consider valid a PR that changes README fragments but doesn't
+          # change the README.rst file itself. It's not really a problem
+          # because the bot will update it anyway after merge. This way, we
+          # lower the barrier for functional contributors that want to fix the
+          # readme fragments, while still letting developers get README
+          # auto-generated (which also helps functionals when using runboat).
+          # DOCS https://pre-commit.com/#temporarily-disabling-hooks
+          SKIP: oca-gen-addon-readme


pedrobaeza:
I agree with this, but I remember that there was a technical dependency for HTML generation. @sbidoul do you remember?

sbidoul:
No, seriously, it's complicated... if we do it in pre-commit we must disable it in the bot because the bot would likely not have the same version of the generator as pre-commit. But old repos don't have pre-commit. So yeah, still not sure how to do that best. Maybe we can enable readme generation in the bot for old versions and use pre-commit for 14+? Or simply abandon readme generation for repos that don't have pre-commit.

yajo:
> I've no idea what you are talking about :)

Probably this: https://github.com/OCA/maintainer-tools/issues/423

> simply abandon readme generation for repos that don't have pre-commit.

That might be a good solution. Less tasks for the bot.

Another one is to instruct the bot to generate the readme using `pre-commit run $new_module oca-gen-addon-readme`.

FWIW [the docutils version is pseudo-pinned](https://github.com/OCA/maintainer-tools/blob/05e217444784fbddf261be1bf1ebb86ad49674c7/setup.py#L31).

fcvalgar:
As a functional reviewer at certain times I find it very difficult to obtain information in the Readme and Usage files which are the fundamental files for us. This is increased when we are reviewing [FIX] [IMP] where there is not much information in these files.

I wish we could have access to the final file before publication.

Thank you contributors

sbidoul:
I have opened https://github.com/OCA/maintainer-tools/issues/564 to discuss the problem and possible solution.

At this stage I think this PR is acceptable in this template, but it should be enabled with a copier question and be disabled by default, because it would not work for OCA.

Shide (approved):
LGTM and speeds up reviewing. We are using this way of work in custom developments and helps a lot. Specially if you want to be careful about readmes

sbidoul (changes requested):
Sorry, I still don't think it is a good idea to enable this by default in OCA.

yajo:
Hello! I've updated the PR to add `--if-fragments-changed`, now that https://github.com/OCA/maintainer-tools/issues/520 is fixed. @sbidoul do you feel more confident now to merge this? Was that flag added too to the bot?

pedrobaeza:
@sbidoul I have noticed also that the current ocabot deployed maintainer-tools seems outdated and it's not generating the links with runboat links and word.

sbidoul:
It's still not clear to me how the bot and a pre-commit hook can/should interact for README generation.
it should be one or the other, otherwise we can't be sure they have the same version.
And a problem with pre-commit is that you have a red build until you run it locally, which makes it impossible tor functional contributors to work on documentation through the GitHub interface.

We need to think all this through, first.

In the meantime, if you want this pre-commit hook, it should be behind a copier question and disabled by default.

yajo:
Hello @sbidoul. I pushed a fix that should resolve the situation. With the new commit, CI will skip this hook. This is a valid usage IMHO because anyways the bot will generate the README if needed. So, if a functional person pushes a change to `readme/DESCRIPTION.rst` and doesn't run pre-commit, the bot will re-generate `README.rst`. But if a dev does that locally and pushes the branch, the `README.rst` will probably be already generated.

Both systems should be configured with `--if-fragments-changed` to avoid conflicts. Here I'm doing it for the hook, but I don't know where to look for the bot configuration to see if it has that flag too.

src/.pre-commit-config.yaml.jinja
```diff
@@ -101,6 +101,12 @@ repos:
       - id: oca-update-pre-commit-excluded-addons
       - id: oca-fix-manifest-website
         args: ["{{ repo_website }}"]
+      - id: oca-gen-addon-readme
+        args:
+          - --addons-dir=.
+          - --branch={{ "%.01f" | format(odoo_version) }}
+          - --repo-name={{ repo_slug }}
+          - --if-fragments-changed
```
sbidoul, line 109:
```suggestion
          - --if-source-changed
```

Also should we add `--no-gen-html` to not generate the `index.html` reduce noise in PRs ?
In any case, the bot will take care of that and do a commit anyway we we bump the version.

Other than that this looks good to me, thanks!

src/.pre-commit-config.yaml.jinja
```diff
@@ -41,7 +41,7 @@
   {%- set repo_rev.flake8 = "3.9.2" %}
   {%- set repo_rev.flake8_bugbear = "21.9.2" %}
   {%- set repo_rev.isort = "5.12.0" %}
-  {%- set repo_rev.maintainer_tools = "4cd2b852214dead80822e93e6749b16f2785b2fe" %}
+  {%- set repo_rev.maintainer_tools = "969238e47c07d0c40573acff81d170f63245d738" %}
```
sbidoul, line 44: At some point the version here and the version in the bot will drift apart, but that is fine I guess.

yajo: If both use `--if-fragments-changed` flag, it shouldn't be a big problem.

At least, as long as the template uses that flag, it should still be OK. If the bot renders the template always, it will harm nobody anyways.

src/.pre-commit-config.yaml.jinja
yajo: I don't think so.

The main issue that this change is trying to fix is the fact that the module README in the runboat doesn't match the fragments when they change.

If we don't render the html, that will still happen, because Odoo displays the html readme if available. So IMHO this should stay as it is.

sbidoul: I hadn't thought about the runboat use case. Makes sense.

sbidoul: Can you just replace with `--if-source-changed` (--if-fragments-changed was confusing because it does not reflect the fact that the manifest is taken into account too).

sbidoul: And a last rebase to resolve the conflict, then good to go as far as I'm concerned.

yajo:
This should be ready to merge.

sbidoul (approved):

sbidoul:
CI ain't happy, though.

legalsylvain:
Hi. I just updated grap repo with latest configuration. (including this PR, that I just discovered)

i fear that this change generates a lot of conflict when there are 2(+) opened PRs against the same module and version. (that occures quite frequently).  I'll take an example: I currently have these 2 PRs that both change the account_invoice_supplierinfo_update module. 
- https://github.com/OCA/account-invoicing/pull/1404/files
- https://github.com/OCA/account-invoicing/pull/1405/files

Today the both PR are compatible, can be gitaggregated together, and can be merge in any order without generating conflict.

But with that changes, (AND the fact that the readme now contain a uniq hash of the files see : https://github.com/OCA/maintainer-tools/commit/3eb1823c905b049a8ca75a42df734b04bd31f6fa), the two PR will now contains a change of the hash that will not be the same. As the result : 
- gitaggregate will fail.
- merging a PR will force the author of the second PR to rebase and fix a conflict on the readme.rst file.

Did I missed something ? 

Thanks !

sbidoul:
I think you are right. Two PRs that both modify the manifest and/or readme fragments of the same addon are going to conflict due to the hash in README.rst.

legalsylvain:
> Two PRs that both modify the manifest and/or readme fragments of the same addon are going to conflict due to the hash in README.rst.

It can occure more frequently. Precisely : conflict will occures if two PRs target the same module, and if the second one changes the manifest and /or readme fragments. 

Rational : whatever the change in the first PR, when we will call ocabot, it will bump the version, so, the manifest will be changed, and the hash will change. 
In that case, a conflict will therefore systematically arise with the 2nd PR.

That's quite annoying. Don't you think ?

legalsylvain:
Thanks for your feedback. I so disabled this feature for my repos. https://github.com/grap/oca-addons-repo-template-new/commit/a7bd24bb6c6cd601fb3996dc36a9c4bbb56aa163

For OCA repos, I see two options : 
- revert the call of oca-gen-addon-readme : https://github.com/OCA/oca-addons-repo-template/pull/220
- have the possibility to call oca-gen-addon-readme with a new option --no-change-hash and use this option in this template. (not sure if it feasible...) what do you think @sbidoul ?

other point of views ? @yajo, @pedrobaeza, @rafaelbn ?

pedrobaeza:
I see very unfrequent that two PRs modify manifest or readme at the same time, so it's not a big deal. I prefer to have this automatically generated.

legalsylvain:
> I see very unfrequent that two PRs modify manifest or readme at the same time, so it's not a big deal. I prefer to have this automatically generated.

As said previously, if a single PR modify readme / manifest it will conflict as soon as another PR changing the same module is merged.

pedrobaeza:
AFAIK, README.rst only changes if there's a change in the contents of the `readme` folder, so the use case you are saying won't conflict.

sbidoul:
The manifest and therefore the version is also part of the files that are hashed.
So I think Sylvain is right.

Either we disable it from pre-commit, or we add a generation mode that keeps the existing hash and we use that in pre-commit, leaving only the bot to update the hash. But that's getting complicated...

pedrobaeza:
Why the manifest/version should be part of the hash if that is not reflected in the README.rst file? I think we should exclude that file from the hash.

sbidoul:
That is possible too but also a brittle hack. The generation mode that keeps the existing hash and we use that in pre-commit would be better I think, but I have not thought about this in details.

yajo:
It is part of the hash because the module name and summary are defined there.

pedrobaeza:
OK, I see. And can this be excluded partially, or is it a whole file hash?

legalsylvain:
> OK, I see. And can this be excluded partially, or is it a whole file hash?

https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py#L519

sbidoul:
> It is part of the hash because the module name and summary are defined there.

And maintainers.

But doing a partial hash excluding the version would be brittle (say, if someone one days adds the version to the template).

legalsylvain:
Hi @yajo. I just used this new configuration that generate an up to date readme file on OCA/16.0 PR.
It does'nt seem to work properly.

- The PR is [here](https://github.com/OCA/pos/pull/1091)
- The autogenerated readme is [here](https://github.com/grap/pos/tree/16.0-ADD-pos_payment_method_change_policy/pos_payment_method_change_policy). 

**First problem**
As you can see, the display of the images are broken because the remote is not ``OCA`` but ``grap`` (in that context) and the branch is not ``16.0`` but ``16.0-ADD-pos_payment_method_change_policy``
As a result, 
- bad url : https://raw.githubusercontent.com/OCA/pos/16.0/pos_payment_method_change_policy/static/description/cash_1_payment_screen.png
- correct url : https://raw.githubusercontent.com/grap/pos/16.0-ADD-pos_payment_method_change_policy/pos_payment_method_change_policy/static/description/cash_1_payment_screen.png 

I see that point as more a regression than an improvment : 
- before : all the images correctly displayed + readme doesn't take into account latest changes (if any)
- now : all the images bad displayed + readme text take into account latest changes (if any).

This will occures : 
- in all migrations PR. (because version is bumped)
- in PR that changes version or change readme fragments

**Second Problem**

In a git point of view, the history of the README.rst file (and index.html file) will be polluted. I mean : 
- commit 1 : by contributor A, break all the images url
- commit 2 : by ocabot, when merging, fix all the images url
- commit 3 : by contributor B, break all the images url again
- commit 4 : by ocabot, when merging, fix all the images url again

Do you know how to fix it ? 

CC : @sbidoul, @pedrobaeza

legalsylvain:
Thanks for your answer ! 

> About problem 1, we have the same problem still today when adding a new module whose readme contains images stored in git.

yes. But the problem is bigger, with the recent change introduced by this PR. If now, a people make a PR against an existing module, changing the readme, all the url of images will be broken, making the readme less understandable. Before it was not the case.

> The solution would be to host those images away from git

I don't know how to do it simply...

yajo:
> I don't know how to do it simply...

You just have to open some github issue and drag and drop the image into the comment box. Github will upload it and give you a URL. Then, remove the image from the git repo and change the URL to that one.

I agree it's not very backwards-compatible. Maybe the convert-to-markdown tool in maintainer-tools could handle this?

simahawk:
FTR

1. FWIW I still think this feature is useless and annoying, sorry :)
2. I've just got again the error 

```
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/bin/oca-gen-addon-readme", line 8, in <module>
    sys.exit(gen_addon_readme())
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/tools/gen_addon_readme.py", line 526, in gen_addon_readme
    gen_one_addon_readme(
  File "/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/tools/gen_addon_readme.py", line 366, in gen_one_addon_readme
    with open(template_filename, "r", encoding="utf8") as tf:
FileNotFoundError: [Errno 2] No such file or directory: '/home/sorsi/.cache/pre-commit/repoipl9kidr/py_env-python3/lib/python3.10/site-packages/tools/gen_addon_readme.rst.jinja'
```
`pre-commit clean` does not help.
Any clue?

However, from now on I will use the workaround that is actually [used by many repo](https://github.com/OCA/web/blob/dada4dc42ee978f04f5988d9c7131d143b07905b/.github/workflows/pre-commit.yml#L38
): set `SKIP: oca-gen-addon-readme` env var while committing to skip this hook.  
Like: `SKIP=oca-gen-addon-readme git commit -m "...whatever..."`

sbidoul:
@simahawk looks similar to https://github.com/acsone/setuptools-odoo/issues :innocent: 

Anyway given the amount of trouble people have with this feature, I'd be inclined to disable it.

pedrobaeza:
Only 2 persons have commented to have troubles about this, Sylvain and Simone, so that's not a lot. And the issue functional people have reviewing PRs with an outdated README is real.

yajo:
... and Simone's error is not related, BTW. It's just an installation problem.

In any case, I agree with @pedrobaeza. For us, this feature has a lot of value. It means more autonomy for the functional members of our team.

However, I'm open to other solutions for the same problem. These come to my mind:
1. Reopen https://github.com/OCA/oca-addons-repo-template/pull/222 where I proposed a fix for your problem.
2. Rebuild READMEs within runboat process, before installing the addons.

PR #198, opened 2023-05-30
build(pylint-odoo): bump for v16

yajo:
Fixes the problem explained in https://github.com/OCA/report-print-send/pull/328#discussion_r1209478993.

PR #196, opened 2023-05-10
[FIX] Make Python version explicit in pre-commit workflow

carmenbianca:
Recently (or recently enough that I've just begun noticing it), the nodeenv dependency had troubles installing. Sometimes. It wasn't very reliably failing.

This is related to <https://github.com/ekalinin/nodeenv/issues/328>.

By explicitly setting the Python version to something <=3.10, we can safely pretend the problem doesn't exist.

Ideally the Python versions match up to the ones used in test.yml, but they don't, and I fear it might be too intrusive to fix that in this commit.

PR #199, opened 2023-06-21
Support Python 2.7

SirTakobi:
>Tests fail because we lost support for Python 2.7
>
>> The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672

_Originally posted by @StefanRijnhart in https://github.com/OCA/oca-addons-repo-template/issues/198#issuecomment-1597034517_

Fix the above issue, and the same issue also in Repositories using this template

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -16,18 +16,21 @@ jobs:
 {%- else %}
     runs-on: ubuntu-22.04
 {%- endif %}
+{%- if odoo_version < 11 %}
+    container:
+      image: python:2.7.18-buster
+{%- endif %}
     steps:
       - uses: actions/checkout@v3
+{%- if odoo_version >= 11 %}
       - uses: actions/setup-python@v2
-{%- if odoo_version < 11 %}
-        with:
-          python-version: "2.7"
-{%- elif odoo_version < 13 %}
+{%- if odoo_version < 13 %}
         with:
           python-version: "3.6"
 {%- elif odoo_version < 14 %}
         with:
           python-version: "3.8"
+{%- endif %}
 {%- endif %}
       - name: Get python version
         run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


SirTakobi:
The repos using the template should be fine with the first commit, but I'm struggling with fixing the CI of the template:

- Python 2.7 seems to be requiring a container, but in there pipx can't be installed because it requires Python3.6.
- I duplicated the steps because didn't manage to use the container conditionally with the matrix configuration.

Any suggestion?

yajo:
Shouldn't we consider dropping py2.7 support too instead?

SirTakobi:
> Shouldn't we consider dropping py2.7 support too instead?

I think this deserves its own issue https://github.com/OCA/oca-addons-repo-template/issues/200

src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -15,19 +15,22 @@ jobs:
     runs-on: ubuntu-20.04
 {%- else %}
     runs-on: ubuntu-22.04
+{%- endif %}
+{%- if odoo_version < 11 %}
+    container:
+      image: python:2.7.18-buster
```
sbidoul, line 21: This may not be necessary since the ubunu-20.04 runner has python2 preinstalled.

src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -15,19 +15,22 @@ jobs:
...
-        with:
-          python-version: "2.7"
-{%- elif odoo_version < 13 %}
+{%- if odoo_version < 13 %}
         with:
           python-version: "3.6"
 {%- elif odoo_version < 14 %}
         with:
           python-version: "3.8"
+{%- endif %}
```
sbidoul, line 33: We should pin the python version for all Odoo versions. But that's for another PR, I guess.

carmenbianca: It is, but the PR was closed.

https://github.com/OCA/oca-addons-repo-template/pull/196

pedrobaeza:
@SirTakobi can you finish this one fixing the CI and attending comments?

sbidoul:
We dropped python 2 support in #217

Issue #195, opened 2023-02-15
[16.0] duplicate-xml-fields should be removed from .pylintrc-mandatory


alexis-via:
I think that, for Odoo v16, the rule "duplicate-xml-fields" should be removed from .pylintrc-mandatory

In a view in v16, if there is a field limited to a group and you're not part of that group, you cannot use that field as filter in a domain.

Let's imagine that you are NOT part of base.group_multi_company. In the view, you have fields with a domain [('company_id', '=', company_id)]. To make it work in v16, you have to do that:
```
<field  name="company_id"  groups="base.group_multi_company"  />
<field name="company_id" invisible="1" />
```

This is the technique used in the official addons. For example, on the sale order form view:
https://github.com/odoo/odoo/blob/16.0/addons/sale/views/sale_order_views.xml#L596
and
https://github.com/odoo/odoo/blob/16.0/addons/sale/views/sale_order_views.xml#L276

But, when you do that, you get this on pre-commit:
```
account_cutoff_base/views/account_cutoff_mapping.xml:20: [W7907(duplicate-xml-fields), ]  Duplicate xml field "company_id" in lines 21
```

So I think we should just disable the rule "duplicate-xml-fields" for v16.

sbidoul:
I moved this issue to oca/oca-addons-repo-template.

legalsylvain:
> Why not using just this 

I'm not in favor of this solution. having to add extra exception when we develop, for valid code is annoying. 

For me, the best solution is what @pedrobaeza propose. (fix the linter). If not possible, I'm in favor to remove it, as @alexis-via proposed.

regards.

yajo:
We can disable the rule for mandatory and leave it enabled for optional pylint checks. This way, it's still informative, and the disabling comment still makes sense, but it also doesn't block appropriate use.

Later, if/when the linter is enhanced, we can enable it again.

StefanRijnhart:
https://github.com/OCA/odoo-pre-commit-hooks/issues/10 seems relevant here. Fixed in pylint-odoo 7.0.5. The check seems to have been removed from the 8.0.0 series entirely: https://github.com/OCA/pylint-odoo/commit/feacd105482859e70e8abf6e594c262d8ee6e1f7#diff-41d180f6f7233d175c2dfe36c45c4ea80eed4754fa7ed4eed46ccde1b2a778c7L31. So any update to a recent version of pylint-odoo should resolve the issue I think.

PR #201, opened 2023-07-03
[IMP] Update pre-commit pylint_odoo version

stefan-tecnativa:
When adding a field with groups and that also belong to a domain, it is necessary to duplicate the field in the view to avoid an error. This was solved for pylint_odoo 7.0.5, but the template is not updated. So updating the version will prevent similar problems with future v16 works.

References:
https://github.com/OCA/odoo-pre-commit-hooks/issues/10
https://github.com/OCA/oca-addons-repo-template/issues/195

ping @pedrobaeza

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -33,7 +33,7 @@
   {%- set repo_rev.prettier = "2.4.1" %}
   {%- set repo_rev.prettier_xml = "1.1.0" %}
   {%- set repo_rev.pylint = "v2.11.1" %}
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Seems OK, but now the CI fails on P2.7 because is not available anymore for GH actions.

yajo:
Could you rebase, please?

pedrobaeza:
Rebased

sbidoul (approved):

pedrobaeza (approved):

yajo (approved):

Issue #200, opened 2023-06-29
Drop Python 2.7


SirTakobi:
Shouldn't we consider dropping py2.7 support too instead?

_Originally posted by @yajo in https://github.com/OCA/oca-addons-repo-template/issues/199#issuecomment-1612626635_

SirTakobi:
For what it's worth, in my company we are still using it so I'd like it to be supported.
On a more personal note, yes it's old and we should no more use it.

carmenbianca:
With my personal hat on: yes, everyone should have dropped Python 2.7 like a hot potato 10 years ago, and it's astonishing that Odoo only started shipping Python 3 support in 2017.

But the character of the OCA is such that everyone kind of works on their own problems in a collaborative fashion, trying not to get in the way of each other. If Python 2.7 support doesn't get in the way of people (and I'm not certain how it would), then I don't see a strong reason to prevent people from working on that.

yajo:
However, we are talking about pre-commit stuff here. You can use pre-commit running on python 3 to lint code written for python 2. And now updating a repo is a imperative action taken by somebody using Copier, so an older repo that would break when being updated can simply not get updated and live happy with its old configurations. Or, when updated, the dev that tries to update it, can just fix the code according to new hooks and keep on being happy too.

sbidoul:
This was done in #217

PR #202, opened 2023-07-11
Allow unsafe `tasks`

SirTakobi:
Otherwise the following message is displayed:
> Template uses unsafe feature: tasks
Introduced in copier v8.0.0 (https://github.com/copier-org/copier/pull/1171)


SirTakobi:
Currently failing until https://github.com/OCA/oca-addons-repo-template/pull/199, there are also things to be decided in https://github.com/OCA/oca-addons-repo-template/issues/200

PR #204, opened 2023-08-12
fix: pylint-odoo v8 configuration options ignored

theangryangel:
Under OCA/pylint-odoo#426, underscores were replaced with dashes for configuration options, when pylint-odoo was bumped in oca-addons-repo-template#198 this change was not taken into account.

I *think* this is the best way to deal with this?

### DIFF
--- a/src/.pylintrc-mandatory.jinja
+++ b/src/.pylintrc-mandatory.jinja
@@ -6,12 +6,21 @@ load-plugins=pylint_odoo
 score=n
 
 [ODOOLINT]
+{%- if odoo_version < 16 %}
 readme_template_url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
 manifest_required_authors={{ org_name }}
 manifest_required_keys=license
 manifest_deprecated_keys=description,active
 license_allowed=AGPL-3,GPL-2,GPL-2 or any later version,GPL-3,GPL-3 or any later version,LGPL-3
 valid_odoo_versions={{ odoo_version }}
+{%- else %}
+readme-template-url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
+manifest-required-authors={{ org_name }}
+manifest-required-keys=license
+manifest-deprecated-keys=description,active
+license-allowed=AGPL-3,GPL-2,GPL-2 or any later version,GPL-3,GPL-3 or any later version,LGPL-3
+valid-odoo-versions={{ odoo_version }}
+{%- endif %}
 
 [MESSAGES CONTROL]
 disable=all
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+import packaging.version
 import yaml
 from copier.main import run_auto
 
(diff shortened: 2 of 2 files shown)


theangryangel:
I’m heading out with family for the afternoon. 

I’ll look at patching the failing test this evening/tomorrow.

pedrobaeza:
Is Odoo v16 using pylint-odoo v8?

theangryangel:
It was included under https://github.com/OCA/oca-addons-repo-template/pull/198/files.
(Assuming im not making the second misread of the week 😅)

pedrobaeza (approved):
One of the tests is red though

theangryangel:
I’ll get the tests adjusted in the morning to account for it 👍

theangryangel:
Hopefully the test changes are appropriate :)

I assume Python 2.7 test failure is to be ignored until #200 is resolved?

pedrobaeza:
Yes, that one is broken for now, but we can manually merge having it red. Merging now, but a new release should be done. Please @sbidoul / @yajo

yajo:
Done: https://github.com/OCA/oca-addons-repo-template/releases/tag/v1.15.1

pedrobaeza:
Thanks!

src/.pylintrc-mandatory.jinja
```diff
@@ -6,12 +6,21 @@ load-plugins=pylint_odoo
...
 license_allowed=AGPL-3,GPL-2,GPL-2 or any later version,GPL-3,GPL-3 or any later version,LGPL-3
 valid_odoo_versions={{ odoo_version }}
+{%- else %}
+readme-template-url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
+manifest-required-authors={{ org_name }}
+manifest-required-keys=license
+manifest-deprecated-keys=description,active
+license-allowed=AGPL-3,GPL-2,GPL-2 or any later version,GPL-3,GPL-3 or any later version,LGPL-3
+valid-odoo-versions={{ odoo_version }}
+{%- endif %}
```
sbidoul, line 23: This solution sounds a bit awkward to me. Summoning @moylop260 here: is this the right way to do this? Did pylint-odoo 8 actually change the option names?

theangryangel: Based on https://github.com/OCA/pylint-odoo/issues/465 thats what I have understood from moylop260 and observed :) If I've made a huge mistake apologies :)

As for the fix I copied the version checking pattern that was being used elsewhere. For changing the tests I explicitly checked the pylint-odoo version. I did toy with doing that, but elected to try and follow the same pattern. I don't mind revisiting this if there's a preferred change.

sbidoul: @theangryangel thanks for the explanation (which I should have found myself, had I read your OP carefully :)

theangryangel: We're all busy people :) It's easily done, anything I can do to try and make life easier tho :)

celm1990:
Hi guys,

I am confused by these changes, for example, I created a [module](https://github.com/celm1990/repo_with_ruff/blob/17.0/my_module/__manifest__.py) with inconsistencies:
- Incorrect version(16.0 instead of 17.0)
- Untranslated text

However, pylint with version ` v8.0.19`(used by this repo for Odoo V17.0) does not fail with it.
image

I tested with `v8.0.5` and now some validations are apllied
image

However `manifest-version-format` does not fail, It is necessary to change dashes to underscores(as before this change), and then it works.
image

So, for pylint-odoo `v8.0.5` these changes are not valid, they must be valid only from v0.8.6+ but it's not working neither (:
image

Can someone please review and provide your comments?

For pylint-odoo v8.0.6+ how should it be configured to work?

theangryangel:
I've had a quick 5 minute check, and agree it is broken, but it's broken specifically on the 17.0 version of the template. 

However, right now I'm equally confused @celm1990 :(

If I take a 16.0 branch, recopy it as 17.0 it breaks. If I put it back to 16.0 it works.

I've got a couple of calls I need to make this morning, but I'll *try* and find the time to diagnose further.

Issue #206, opened 2023-09-07
Error: No such option: --if-source-changed


zamberjo:
## Describe the bug
After updating a repository (e.g. OCA/l10n-spain/tree/15.0) an error appears in the hook "oca-gen-addon-readme".

```bash
Usage: oca-gen-addon-readme [OPTIONS]
Try 'oca-gen-addon-readme --help' for help.

Error: No such option: --if-source-changed
```

It seems that revisions ["ab1d7f6" (<15)](https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pre-commit-config.yaml.jinja#L12C8-L12C8) and ["dfba427ba03900b69e0a7f2c65890dc48921d36a" (<16)](https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pre-commit-config.yaml.jinja#L29) do not have such an option. While ["969238e47c07d0c40573acff81d170f63245d738"](https://github.com/OCA/oca-addons-repo-template/blob/master/src/.pre-commit-config.yaml.jinja#L46) does.

## To Reproduce
Update the template in any repository in version 15.0 - 16.0

yajo:
> Update the template in any repository in version 15.0 - 16.0

16.0 works fine: https://github.com/OCA/web/pull/2609

It's the rest of branches that fail.

PR #182, opened 2022-12-15
Pin python version for <= 13.0

carmenbianca:
Fixes #174

PR #207, opened 2023-09-07
fix: incremental addon readme generation for v15 and older too

yajo:
Fixes https://github.com/OCA/oca-addons-repo-template/issues/206


pedrobaeza (approved):

zamberjo (approved):

pedrobaeza:
One question: that maintainer-tools revision is compatible with all the Python versions in lower Odoo versions?

pedrobaeza:
Not really. It was adapted for 12 and 13 as well. Example: https://github.com/OCA/oca-addons-repo-template/pull/182

PR #210, opened 2023-09-12
Generate README for selected Organization

SirAionTech:
Before this change, links in the README are always generated as `OCA/repo_slug` despite the user may have answered a different organization to the question
> What's the organization slug? If you are creating https://github.com/OCA/server-tools, then write "OCA" here.


SirAionTech:
Tests fail probably due to https://github.com/ekalinin/nodeenv/issues/324.

legalsylvain (approved):
thanks !

PR #211, opened 2023-09-12
Don't bother non-technical authors with minor formatting issues in readme fragments

sbidoul:
I propose to exclude `readme/` fragments from linters.

As long as the `README.rst` generator accepts them and produces valid `.rst` (which is checked upon `ocabot merge`), we should not bother non technical authors with overly strict formatting issues of readme fragments.

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -77,6 +77,8 @@ exclude: |
   /static/(src/)?lib/|
   # Repos using Sphinx to generate docs don't need prettying
   ^docs/_templates/.*\.html$|
+  # Don't bother non-technical authors with formatting issues in docs
+  readme/.*\.(rst|md)$|
   # You don't usually want a bot to modify your legal texts
   (LICENSE.*|COPYING.*)
 default_language_version:


pedrobaeza:
But AFAIK, this will make that until merge, there's no way to detect possible problems. I think the same should apply for both technical and non technical...

sbidoul:
> But AFAIK, this will make that until merge, there's no way to detect possible problems. I think the same should apply for both technical and non technical...

Not really. The "issues" pre-commit "fixes" in markdown text are not important. Mostly trailing whitespace, line wrapping etc. They do not detect markdown validity errors. We could have a dry-run generation of the README in CI, but that is a different issue.

Look at the linked PR as an example of what pre-commit complains about.

So I think it is more important to facilitate docs contributions by non-tech folks than being pedantic about whitespace in documentation.

pedrobaeza:
OK then.

yajo (approved):
The bad part about this is that, when a dev comes later and edits those docs, chances are that some autoformatter will reformat it or, at least, remove those whitespaces.

That's uncomfortable, although probably the lesser bad. I understand your reasoning.

carmenbianca (approved):
ack @yajo 's comments

sbidoul:
Thanks for the valuable feedback everyone!

@tumbaoju FYI.

PR #215, opened 2023-10-28
Ignore build and dist directories in linters

sbidoul:
These directories are typically the output of python build tools. The are already ignored in `.gitignore`.


pedrobaeza (approved):
Why don't we remove the P2.7 test already for having green CI?

rvalyi (approved):

yajo:
Do we really need to ignore them here if they are ignored in `.gitignore`? Doesn't pre-commit already respect `.gitignore` automatically? 🤔

sbidoul:
I did not know pre-commit did that. But I did have pylint-odoo problems without this.

PR #216, opened 2023-10-28
Add pyproject.toml instead of setup.py

sbidoul:
Add a question to generate pyproject.toml files instead of setup.py.

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -112,6 +112,12 @@ dependency_installation_mode:
   help: Choose how to install module dependencies.
   when: *ci_is_travis
 
+use_pyproject_toml:
+  default: no
+  type: bool
+  help: Use pyproject.toml instead of setup.py.
+  when: "{{ odoo_version >= 14.0 }}"
+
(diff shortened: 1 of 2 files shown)


yajo:
**thought:** Couldn't we just add this always? IIUC, having a `pyproject.toml` file based on whool inside the addon dir, and having a `setup/` folder based on setuptools-odoo, shouldn't conflict, right? Whool would just be the new official way to install addons, and setuptools-odoo would be deprecated and exist only on older branches.

carmenbianca:
I wouldn't do both simultaneously in any case, even if they don't conflict. A single source of truth is more important, and it's _far_ easier to debug problems if you don't also have to divine which package installation method is being used.

But I've not properly reviewed the code, so I may change my mind.

sbidoul:
Yes I chose to have one OR the other because they are sometimes modified manually and we don't want to tell people to modify both in that case.

etobella (approved):

PR #218, opened 2023-10-28
Simplify linters using ruff

sbidoul:
Simplify the linters of *this project* (not the templatate... yet).


sbidoul:
I'm going ahead and merge this 
- there is no risk as this only lints the tests of this project, basically
- the fact that the flake8 and isort config are symlinks to the template configs is problematic as this triggers surprising side effects so since this PR removes them, it simplifies maintenance of this repo

PR #219, opened 2023-10-28
Use ruff instead of black, isort, flake8, autoflake, pyupgrade

sbidoul:
From the [docs](https://docs.astral.sh/ruff/): *An extremely fast Python linter and code formatter, written in Rust.*.

The goal is multiple
- better pre-commit performance
- less dependencies
- ruff being a static executable it will be less senstive to python upgrades
- basic set of rules enabled by default, a copier question let PSC add additional rules if they want to

I tried to configure rules that are compatible with the current flake8 and isort config.

To test this on a repo, run 

`pipx run copier copy git+https://github.com/acsone/oca-addons-repo-template --vcs-ref=use-ruff --trust .`

and answer `yes` to the "use ruff" question.


sbidoul:
Here is an example of what this PR does: https://github.com/OCA/mis-builder/pull/577

I would not enable this by default on existing branches but I think it's ok for 17.0.

What do people think?

pedrobaeza:
OK, let's go with it by default in v17.

copier.yml
```diff
@@ -126,6 +126,18 @@ generate_requirements_txt:
     Generate requirements.txt from addons manifests and optional overrides in setup.py
     files.
 
+use_ruff:
+  default: no
```
yajo, line 130: Since this requires a manual template upgrade and ruff is a better default, I think we should enable it by default.

```suggestion
  default: yes
```

copier.yml
```diff
@@ -126,6 +126,18 @@ generate_requirements_txt:
...
+  type: bool
+  help: Use ruff and ruff-format instead of flake8, autoflake, pyupgrade, isort, black.
+  when: "{{ odoo_version >= 14.0 }}"
+
+additional_ruff_rules:
+  type: yaml
+  default: []
+  help: List of additional ruff rules to enable.
+  when: "{{ use_ruff }}"
+
```
yajo, line 140: I think you don't need to add this here.

Copier supports smart updates. Thus, if you need extra rules, all you have to do is edit the ruff config file. Changes will be preserved.

```suggestion
```

src/.pre-commit-config.yaml.jinja
```diff
@@ -176,11 +178,14 @@ repos:
       - id: check-xml
       - id: mixed-line-ending
         args: ["--fix=lf"]
+  {%- if not use_ruff or odoo_version < 15.0 %}  # ruff doesn't support python 3.6
```
yajo, line 181: AFAICS in the copier questionary, you cannot use ruff in Odoo versions lower than 15, so this or is redundant:

```suggestion
  {%- if not use_ruff %}  # ruff doesn't support python 3.6
```

copier.yml
```diff
@@ -126,6 +126,18 @@ generate_requirements_txt:
     Generate requirements.txt from addons manifests and optional overrides in setup.py
     files.
 
+use_ruff:
+  default: no
+  type: bool
+  help: Use ruff and ruff-format instead of flake8, autoflake, pyupgrade, isort, black.
+  when: "{{ odoo_version >= 14.0 }}"
```
yajo, line 133: Upgrade the minimal Copier version to make sure `when` and `default` work reliably.

src/.pre-commit-config.yaml.jinja
```diff
@@ -201,12 +207,22 @@ repos:
...
       {%- endif %}
+  {%- if not use_ruff %}
   - repo: https://github.com/PyCQA/flake8
     rev: {{ repo_rev.flake8 }}
     hooks:
       - id: flake8
         name: flake8
         additional_dependencies: ["flake8-bugbear=={{ repo_rev.flake8_bugbear }}"]
+  {%- endif %}
+  {%- if use_ruff %}
```
yajo, line 218:
```suggestion
  {%- else %}
```

src/.pre-commit-config.yaml.jinja
```diff
@@ -176,11 +178,14 @@ repos:
...
       - id: mixed-line-ending
         args: ["--fix=lf"]
+  {%- if not use_ruff or odoo_version < 15.0 %}  # ruff doesn't support python 3.6
   - repo: https://github.com/asottile/pyupgrade
     rev: {{ repo_rev.pyupgrade }}
     hooks:
       - id: pyupgrade
         args: ["--keep-percent-format"]
+  {%- endif %}
+  {%- if not use_ruff %}
```
yajo, line 188: Following the suggestion above:

```suggestion
```

src/{% if use_ruff %}.ruff.toml{% endif%}.jinja
```diff
@@ -0,0 +1,37 @@
...
+extend-select = [
+    "B",
+    "C90",
+    "I",  # isort
+    {%- if odoo_version >= 15.0 %}
+    "UP",  # pyupgrade
+    {%- endif %}
+    {%- for rule in additional_ruff_rules %}
+    "{{ rule }}",
+    {%- endfor %}
```
yajo, line 18: Following suggestions above:

```suggestion
```

yajo (changes requested):

copier.yml
sbidoul: Would that not enable ruff by default when we apply the template on existing repos?

copier.yml
sbidoul: In general I find that brittle with conflicts on updates. When we have question it is more robust.

src/.pre-commit-config.yaml.jinja
sbidoul: The copier question accepts ruff for 14.

copier.yml
legalsylvain, line 130:
```suggestion
  default: |-
    {% if odoo_version < 17.0 -%}
        no
    {%- else -%}
        yes
    {%- endif %}
```

?

sbidoul: Good idea. Done.

copier.yml
sbidoul: @yajo I set min_copier_version to 9. Is that ok or do you know a better value?

sbidoul:
I have handled the review comments and replied to some. @yajo can you update your review?

copier.yml
sbidoul: Hm that does not work because the tests here seem to be incompatible with copier 9. They should be rewritten to use the CLI I think. That will be for another day. Help welcome.

yajo: https://github.com/OCA/oca-addons-repo-template/pull/221

copier.yml
yajo: OK, nevermind. There are more than 1 way to do things right. 😉

src/.pre-commit-config.yaml.jinja
yajo: Ah true.

yajo (approved):
Thanks!

copier.yml
sbidoul: And other argument in favor of the question approach, is that we pick-up answers from the previous branches when creating new ones.

PR #221, opened 2023-10-31
build: support copier 9

yajo:


copier.yml
```diff
@@ -1,14 +1,7 @@
...
 _templates_suffix: .jinja
 _subdirectory: src
-_envops:
-  block_end_string: "%}"
-  block_start_string: "{%"
-  comment_end_string: "#}"
-  comment_start_string: "{#"
-  keep_trailing_newline: true
-  variable_end_string: "}}"
-  variable_start_string: "{{"
```
sbidoul, line 11: @yajo does this mean the default jinja syntax changed with copier 9?

sbidoul (approved):

copier.yml
yajo: It changed many versions ago. We just didn't have a constraint, so we had to define it manually.

pedrobaeza:
Why forcing the latest version and why each version is incompatible with the previous?

yajo:
There are a few fixes for `default` and `when` in recent versions. I can't remember in which one there were some bugs fixed, but anyway it seems sane to bump to latest.

Copier is detached from the template, so ensuring a minimal version is used means more predictable replayability.

Regarding incompatibilities, you can see [the changelog](https://copier.readthedocs.io/en/stable/changelog/).

pedrobaeza:
My fear here is that copier won't run in its latest version in all current development environments. Is it working for current + previous LTS versions of popular distros?

yajo:
AFAIK no popular distro packages copier.

yajo:
If you install copier with pipx or nix, it won't be a problem.

pedrobaeza:
OK, good to know. I'll write back if any problem found.

PR #220, opened 2023-10-29
Revert "feat: generate addons readme in pre-commit"

legalsylvain:
This reverts commit a1551e5d312fc8c48571b124e02c91c2c19d671a.

Rational: The call of oca-gen-addon-readme will generate conflicts in the readme.rst file when 2 PRs modify the same module. See https://github.com/OCA/oca-addons-repo-template/pull/190#issuecomment-1784116516

PR #222, opened 2023-10-31
fix(oca-gen-addon-readme): reduce unneeded merge conflicts

yajo:
As seen in https://github.com/OCA/oca-addons-repo-template/pull/190#issuecomment-1784076360, when somebody tried to merge 2 PRs for the same module, chances are that both imply different readme digests and the merge fails.

As seen in https://github.com/OCA/oca-addons-repo-template/pull/220#issuecomment-1786814130, this was already happening under some circumstances with `requirements.txt` files.

Here I added a `.gitattributes` file which specifies a `merge=union` driver for those files. [Git docs][1] explain that this driver will just keep the changes from all files, marking the merge as succeeded and not including merge markers. This is not ideal, but is OK for our current situation because:

1. Duplicate/conflicting lines in those files should not really harm a lot.
2. The bot will re-render the README upon merge anyways, fixing that conflict automatically.
3. We could define a custom merge driver, but `union` is built into Git, meaning less configurations required downstream.

Still, it doesn't seem to work in octopus merges. At least, if you do consecutive merges instead, all will work as expected.

At the same time, I moved the hook to be the last. This is because previous hooks can reformat the files used to obtain the digest.

Closes #220.

[1]: https://git-scm.com/docs/gitattributes.html#Documentation/gitattributes.txt-union

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,17 @@ jobs:
       - name: Install poetry
         run: pipx install poetry
 
+      # Some tests exercise pre-commit; all use poetry; cache that
+      - name: Restore poetry and pre-commit cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pre-commit
+            ~/.cache/pypoetry/virtualenvs
+          key:
+            test-${{ runner.os }}-${{ hashFiles('poetry.lock',
+            '**/.pre-commit-config.yaml.jinja') }}
+
       # Install specific alternate python versions required by hooks
       - name: Install python
         uses: actions/setup-python@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,32 +9,9 @@ on:
 
 jobs:
   test:
-    runs-on: ${{matrix.machine}}
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
-      matrix:
-        include:
-          - odoo-version: 11.0
-            python-version: 3.6
-            machine: ubuntu-20.04
-          - odoo-version: 12.0
-            python-version: 3.6
-            machine: ubuntu-20.04
-          - odoo-version: 13.0
-            python-version: 3.8
-            machine: ubuntu-22.04
-          - odoo-version: 14.0
-            python-version: 3.8
-            machine: ubuntu-22.04
-          - odoo-version: 15.0
-            python-version: 3.8
-            machine: ubuntu-22.04
-          - odoo-version: 16.0
-            python-version: "3.10"
-            machine: ubuntu-22.04
-          - odoo-version: 17.0
-            python-version: "3.10"
-            machine: ubuntu-22.04
     steps:
       # Prepare environment
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


src/.pre-commit-config.yaml.jinja
```diff
@@ -235,3 +229,12 @@ repos:
         args:
           - --rcfile=.pylintrc-mandatory
 {%- endif %}
+  - <<: *maintainer_tools
```
legalsylvain, line 232: not sure to understand this syntax.

legalsylvain:
Thanks a lot for your investigation and your PR ! I'll try this code in the oca days.

.github/workflows/test.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - run: git config --global user.email CI@GITHUB
 
       # Run all tests
-      - run: poetry install
+      - run: poetry install --with dev
```
pedrobaeza, line 53: Why this change?

src/.pre-commit-config.yaml.jinja
yajo: It's reusing the anchor defined above. Check this out: https://www.linode.com/docs/guides/yaml-anchors-aliases-overrides-extensions/

.github/workflows/test.yml
yajo: The CI failures for previous commit are weird. I don't get those testing locally. I was wondering if maybe pre-commit or pyyaml were too old. Or maybe if a more modern poetry isn't installing dev dependencies, as older poetry did. They had quite a big change in [dependency groups management](https://python-poetry.org/docs/managing-dependencies#dependency-groups).

yajo:
Hi there. This one is ready to merge.

I was having problems because the tests were run under python 3.8 but pre-commit needed python 3.6. I had to pin an older virtualenv version because the latest one doesn't support python 3.6 anymore.

Long story short, now we don't issue a matrix of builds. Instead, we install all required python interpreters in a single build and parallelize at pytest level with pytest-xdist. Thus, now there's just one "test" job.

Since there's just that job, the CI checks will stay as "Required statuses must pass before merging" because I changed the way CI is handled here.

If nobody's against this, I'll merge it so we can unblock the situation explained at https://github.com/OCA/oca-addons-repo-template/pull/220#issuecomment-1802291979.

pedrobaeza (approved):

sbidoul:
@legalsylvain have you tested this?

sbidoul:
I'm proposing using https://github.com/OCA/maintainer-tools/pull/588 as an alternative.

yajo:
I still think this is a better solution.

Multiple PRs can modify instructions in the readme folder. If those conflict, there's no other solution than to solve it. But if the conflict happens in the README, we shouldn't care when merging, as the bot would re-trigger the generation after merge.

Also this PR refactors the test suite in a more future-proof way (that can be extracted separately, of course).

So IMHO this still has its benefits and we should merge this.

sbidoul:
> Also this PR refactors the test suite in a more future-proof way (that can be extracted separately, of course).

:+1: for that

Regarding the union merge, I think it may be fine for requirements.txt (although I have not personally witnessed such conflicts in practice).

For README.rst/index.hml, I'm not sure.  The drawbacks you mention in the OP don't seem to be worth the gain.

yajo:
OK, nevermind. We can come back to this later if needed.

PR #223, opened 2023-11-01
[FIX] Only execute the PO file review for files inside a 18n folder

etobella:
Applies the fix for https://github.com/OCA/odoo-pre-commit-hooks/pull/90

tested on https://github.com/OCA/connector/pull/469

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -120,6 +120,7 @@ repos:
     hooks:
       - id: oca-checks-odoo-module
       - id: oca-checks-po
+        files: '/i18n/[a-zA-Z0-9_]*\.(po|pot)$'
   {%- if not use_ruff %}
   - repo: https://github.com/myint/autoflake
     rev: {{ repo_rev.autoflake }}


src/.pre-commit-config.yaml.jinja
```diff
@@ -120,6 +120,7 @@ repos:
     hooks:
       - id: oca-checks-odoo-module
       - id: oca-checks-po
+        files: '/i18n/[a-zA-Z0-9_]*\.(po|pot)$'
```
moylop260, line 123: IMHO we shouldn't define a particular regex to filter the file because we could miss a case

e.g. `module-name.pot`

Notice the dash `-` for this file
 
What about?

```suggestion
        files: '/i18n/*\.(po|pot)$'
```

sbidoul: But a module name cannot contain a dash?

moylop260: @sbidoul 

Oh, you are right, as usual :)


@etobella 

IMHO it should be in the hooks repo:
  - https://github.com/OCA/odoo-pre-commit-hooks/blob/5e97e13778f2cf7440bc711171f9e4db69a0f6a7/.pre-commit-hooks.yaml#L24

etobella: Well, I made the fixing PR in both repos.

IMO is cleaner to edit the precommit, as it only affects a little set of repositories with documentation

yajo (changes requested):
I agree with https://github.com/OCA/oca-addons-repo-template/pull/223#discussion_r1379346584. This one doesn't seem to be the right place to fix this. It should be fixed upstream and then, here, we just include a newer version of that repo.

PR #231, opened 2023-11-12
Enable ruff line too long check

sbidoul:
It is disabled by default because ruff-format makes a best effort attempt to satisfy it and may fail otherwise.

It should fine to enable it as it was with flake8 (the default is 88).

See https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules for more information.


celm1990 (approved):

celm1990:
@sbidoul 
Do you think this change can be solve this log on this PR https://github.com/Tecnativa/doodba-copier-template/pull/425
image

pedrobaeza (approved):

PR #235, opened 2023-11-17
[IMP] upgrade pylint version to 9.0.4 to support Odoo 17.0

celm1990:
see https://github.com/OCA/pylint-odoo/pull/477

@pedrobaeza @sbidoul could you please review this?

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -54,6 +54,10 @@
   {%- set repo_rev.setuptools_odoo = "3.1.8" %}
 {%- endif %}
 
+{%- if odoo_version > 16 %}
+  {%- set repo_rev.pylint_odoo = "v9.0.4" %}
+{%- endif %}
+
 {#- Older versions that differ a lot have their own hardcoded templates for readability #}
 {%- if odoo_version < 13 %}
   {%- include "version-specific/mqt-compat/.pre-commit-config.yaml.jinja" %}


pedrobaeza (approved):

sbidoul:
This change applies to 16 too. Do we risk making existing 16.0 branches red due to this?

celm1990:
> This change applies to 16 too. Do we risk making existing 16.0 branches red due to this?

@sbidoul but for 16 it is working, because 16.0 already is into `DFTL_VALID_ODOO_VERSIONS` https://github.com/OCA/pylint-odoo/blob/d5888089eac660ebb6c06f33e3ed5655c409eb72/src/pylint_odoo/misc.py#L19-L33

celm1990:
@sbidoul, my apologies for the confusion. I have revised the code to align with your reference.

yajo:
> Do we risk making existing 16.0 branches red due to this?

Is this question really relevant today?

In old times, when linting was centralized in MQT, a single update there could set many branches red.

Now with the copier template, you only break things _if_ and _when_ you update the template in a branch. At that moment in time, you have the chance to apply auto or manual fixes, and then push the new green branch that satisfies any new rules.

So, do we really need to care that much now about backwards compatibility? 🤔

sbidoul:
> Is this question really relevant today?

I think it is. Sometimes we need to do a mass update (like we recently did for 14, 15, 16, which caused a lot of troubles).

I think we must reach a situation where the list of lint checks is explicit and not dependent of the specific version of the linters.

celm1990:
@sbidoul  Can we get this merged, at least for new repositories (from OCA or personal ones)? I understand that a mass update to existing OCA repositories may not necessarily occur. Is there anything we can improve in the pull request (PR) to facilitate the merge?

PR #236, opened 2023-11-18
[FIX] Do not apply XML new style checks to 14.0

dreispt:
This will about most of the failing updates to 14.0 dotfiles, since the current pre-commit is running new checks not verified on the previous pre-commit version.

Example where this was verified and got a green build:
https://github.com/OCA/product-attribute/pull/1440

cc @etobella @sbidoul @moylop260


sbidoul:
Looks good to me.

PR #240, opened 2023-12-11
Ignore test files in linters

tuantrantg:


src/.pre-commit-config.yaml.jinja
```diff
@@ -85,6 +85,8 @@ exclude: |
   readme/.*\.(rst|md)$|
   # Ignore build and dist directories in addons
   /build/|/dist/|
+  # Ignore test files in addons
+  /tests/.*\.(wamas|txt|xml)$|
```
simahawk, line 89:
```suggestion
  /tests/fixtures/*|
```
I would ignore everything in a given dir.
If by chance someone wants to have formatted xmls there the it will be their choices.
Also, I would not tie this to any specific extension.

yajo (changes requested):
Could you please explain the motivation behind this PR?

src/.pre-commit-config.yaml.jinja
tuantrantg: hi @simahawk the PR has been updated as your request

tuantrantg:
> Could you please explain the motivation behind this PR?

hi @yajo thanks for asking. I'm doing some modules that need XML / text files to run unitest and `pre-commit` beautifies XML / text files and I want to prevent to do this

simahawk:
> Could you please explain the motivation behind this PR?

Example: you get an example file (xml maybe) from an official doc/example and you want to respect the original. There's no reason to re-format it. 
And it's optional in any case.

src/.pre-commit-config.yaml.jinja
```diff
@@ -85,6 +85,8 @@ exclude: |
   readme/.*\.(rst|md)$|
   # Ignore build and dist directories in addons
   /build/|/dist/|
+  # Ignore test files in addons
+  /tests/fixtures/*|
```
yajo, line 89: Thanks for the explanations.

I think "fixtures" is very confusing here, as it has no special meaning, and otherwise it reminds to pytest fixtures, which have no relationship with what you want.

What about this?

```suggestion
  /tests/vendor/*|
```

"Vendor" is a common word for external code or files that are included in an alien repo.

simahawk: Fine for me. @tuantrantg can you update?

carmenbianca: I would recommend `/tests/resources` as a more generic name for files which the tests depend on. This directory needn't necessarily contain 3rd-party stuff. Failing that, `/tests/vendor` is also fine.

pedrobaeza: Or even `samples` better than `vendor` one IMO.

tuantrantg:
hello @pedrobaeza @carmenbianca @simahawk @yajo I finally updated this to `/tests/samples`

pedrobaeza (approved):

simahawk (approved):

simahawk:
@yajo fine for you?

yajo (approved):

Issue #244, opened 2024-01-18
[RFC] Default odoo_test_flavor for recent version 


legalsylvain:
Hi all,

### Context
 
I [recently updated](https://github.com/OCA/pos/pull/1134) the copier answer of the repo OCA/pos for branch 16.0. In that occasion I wanted to switch odoo_test_flavor to "Odoo". @pedrobaeza said me that it was not a valid option so I reverted that changes. 

### Rational
The rational behind the proposal was the following : 

- CI consume energy and it would be desirable to save it, _if possible_.
- if we have to choose, OCA modules _should_ work with odoo/odoo because there are a lot of instances working on that version. OCA/OCB is very confidential.
- The paradigm to test PR with OCB is maybe not the good point. I mean: 
    - if a PR doesn't work with odoo/odoo, it should be fixed.
    - but if a PR works with odoo/odoo and not with OCA/OCB, it should not be blocking for sure for the PR, and maybe it is OCB that should be fixed.
- But above all, I see OCB as an a project at the end of life. Since V13+ there are less that 10 commits per version and most of the commit are trivial. (changing some auto_install to False that could be done via other OCA recent solution with odoo/odoo and [module_change_auto_install](https://github.com/OCA/server-tools/tree/16.0/module_change_auto_install)).

![image](https://github.com/OCA/oca-addons-repo-template/assets/3407482/0684cbac-7cc6-4fff-b042-153349ff2ca4)

[OCB Evolution of PR.xlsx](https://github.com/OCA/oca-addons-repo-template/files/13978813/OCB.Evolution.of.PR.xlsx)

### Proposal

- In recent version, (15 / 16 / 17) : set default to "Odoo" in the template
- allow to let OCB, to allow to test in both context for very specific cases, like the l10n_fr_oca fork done by @alexis-via. (Ref : https://github.com/OCA/OCB/pull/1138 / https://github.com/OCA/l10n-france/issues/379)

CC : 
- funder :  @StefanRijnhart 
- commiters that makes changes in OCB on the recent versions (14 / 15 / 16) :  @victoralmau, @chienandalu, @remi-filament, @alexis-via, @igallart, @MiquelRForgeFlow, @joao-p-marques

thank you for reading.

yajo:
One of OCB rules has always been: keep it 100% compatible with upstream Odoo. Thus, the amount of changes that can really land in OCB is extremely low.

[...]

PR #245, opened 2024-01-30
[IMP] Propose Odoo as default version for recent version to avoid useless consumption of CI

legalsylvain:
Fix: #244

CC : @yajo, @sbidoul

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -76,7 +76,7 @@ odoo_test_flavor:
     {% if odoo_version < 13.0 -%}
         OCB
     {%- else -%}
-        Both
+        Odoo
     {%- endif %}
   choices:
     - Odoo


pedrobaeza:
https://github.com/OCA/oca-addons-repo-template/issues/244#issuecomment-1917383907

yajo (approved):
FWIW this is just changing the default, which means that all repos will stay as they are now unless manually changed.

OTOH, whenever we branch off 18.0, these will default to only use Odoo, which makes sense because there'll be no known difference with OCB at the day of launch.

In general I agree with this, as explained in the issue. Thanks!

carmenbianca (approved):
Agree that electricity savings outweigh the borderline needs of testing against both all the time.

Odoo (instead of OCB) is a good default, but I could be convinced both ways.

If we absolutely must test against both versions, we should exclusively do so on the main branch, not on the PRs, and also not on the ocabot branch.

Issue #250, opened 2024-04-30
Codecov blocking CI jobs


sbidoul:
Since yesterday, the codecov action step is blocking all our CI jobs.

This blocks basically everything as there is a timeout of several hours on GitHub actions.

Issue submitted to codecov here: https://github.com/codecov/feedback/issues/354

Things to try

1. find a way to cancel all queued jobs, so we can try new things without waiting for the job timeouts
2. try if the latest codecov action (v4) works
  - if not, remove the codecov step from this template...
  - if it works, update this template

rousseldenis:
I think they removed their v1 actions:

![image](https://github.com/OCA/oca-addons-repo-template/assets/19529533/8be976d1-b33f-4f8e-ab11-b3a03cc9d151)

IMHO, we can switch all the repos to v3

sbidoul:
Now cancelling all queued jobs:

```python
import json
import subprocess


def get_repos(org: str):
    res = subprocess.check_output(
        ["gh", "repo", "list", "--limit", "1000", "--json", "name", org]
    )
    for repo in json.loads(res):
        yield f"{org}/{repo['name']}"


def get_queued_jobs(repo: str):
    res = subprocess.check_output(
        [
            "gh",
            "run",
            "--repo",
            repo,
            "list",
            "--status",
            "queued",
            "--json",
            "databaseId",
        ]
    )
    for job in json.loads(res):
        yield job["databaseId"]


for repo in get_repos("OCA"):
    print(repo)
    for job in get_queued_jobs(repo):
        subprocess.run(["gh", "run", "--repo", repo, "cancel", str(job)])
 ```

yajo:
> v4 requires TOKEN configuration

Can this be done adding a organization-wide secret? In such case, I guess it's better to update to the newest version, now that we do it.

sbidoul:
> Can this be done adding a organization-wide secret?

Yes absolutely. I have already added an org wide secret named `CODECOV_TOKEN`.

See example here: https://github.com/OCA/mis-builder/pull/606

This does not block anymore but fails to find the coverage report, though.

rousseldenis:
> If anyone wants to submit a PR to upgrade to codecov-action v4, under a copier question, that would be helpful.

I'm on it

SirAionTech:
> > Can this be done adding a organization-wide secret?
> 
> Yes absolutely. I have already added an org wide secret named `CODECOV_TOKEN`.
> 
[...]

PR #251, opened 2024-04-30
[IMP] Add codecov v4 question

rousseldenis:
Solves:

- https://github.com/OCA/oca-addons-repo-template/issues/250

### DIFF
--- a/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
+++ b/src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
@@ -145,7 +145,9 @@ jobs:
       - name: Run tests
         run: oca_run_tests
       {%- if github_enable_codecov %}
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@v4
+        with:
++          token: {{"${{ secrets.CODECOV_TOKEN }}"}}
       {%- endif %}
       {% raw -%}
       - name: Update .pot files


copier.yml
```diff
@@ -175,6 +175,12 @@ github_enable_codecov:
   help: GitHub action runs codecov/codecov-action?
   when: *ci_is_github
 
+github_v4_codecov:
```
pedrobaeza, line 178: Should this be a question? I think it should be the only option.

rousseldenis: I can, this is simpler. @sbidoul

sbidoul: Since we already have github_enable_codecov, I don't think a new question is necessary.

rousseldenis: Ok, that leads to this change :smile: :

image

sbidoul: Yeah I had not checked, so I did not know we had a question for codecov already.

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -145,7 +145,8 @@ jobs:
       - name: Run tests
         run: oca_run_tests
       {%- if github_enable_codecov %}
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@v4
+      {%- endif %}
```
pedrobaeza, line 149: This one should go out, isn't it?

rousseldenis:
This one is ready

pedrobaeza (approved):
The only pending thing is if the token is needed or not, but I suppose it not too much burden to add such token globally in the organization, so better to have it at once and avoid future problems.

celm1990:
Hi, I use Codecov for private repositories, but with `codecov-action V4`, it doesn't work with PRs from `forks` because tokenless usage is unsupported.

image

Additionally, GitHub Actions does not send secrets from forks.
https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

image

According to this [link](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories
), I enabled these configurations in my repository's settings, but it's still not working.

image


As a workaround, for now, I am using `V3.1.6`, and it works. It's not the best practice, but do you have any other solutions?

PR #253, opened 2024-05-02
fix(codecov): continue on error

yajo:
This should help in the future to protect against situations like https://github.com/OCA/oca-addons-repo-template/issues/250, when Codecov suddenly changes something, starts to fail and becomes blocking for everybody here.


sbidoul:
I think that would not have helped in the current situation because the codecov step does not fail, it just blocks.

legalsylvain (approved):
Make sense.

yajo:
OK, but it would help in case codecov server goes down, for example. That'd be mostly as dramatic as #250.

PR #256, opened 2024-05-07
[FIX] 15/16/17: Removed first line break on pre-commit template

manuel-florido:
Hello,

When using v15/v16/v17, It creates .pre-commit-config.yaml with a line break on the first line. 

If you use pre-commit It fails at the first attempt because this line gets removed.

With this change, the file is created correctly

Thank you!


legalsylvain (approved):
Thanks !

zamberjo (approved):

carmenbianca (approved):
LGTM, thanks

Issue #258, opened 2024-05-22
ESlint new configuration file format.


PieterPaulussen:
**Is your feature request related to a problem?**
Yes, the ESLint configuration file has been refactored into a new format. See [here](https://eslint.org/docs/latest/use/configure/configuration-files). From version 9.3.0, the old format is no longer supported.

**Describe the solution you'd like**
The template ESLint file should be refactored / converted into the new ESLint 9.3.0 format.

**Describe alternatives you've considered**
None

sbidoul:
This has been done 18.0 branches.

PR #259, opened 2024-05-22
[FIX] Support ecmaVersion 2022

chienandalu:
Odoo officially supports ecmaVersion 2022 from v17: https://github.com/odoo/odoo/blame/9a11717c17b860ec2f1b2e228517c0d3945c474a/addons/web/tooling/_eslintrc.json#L9

Possible fixes:

- Support it globally
- Make an special template for Odoo >= 17

What do you think?

### DIFF
--- a/src/{% if odoo_version > 12 %}.eslintrc.yml{% endif %}.jinja
+++ b/src/{% if odoo_version > 12 %}.eslintrc.yml{% endif %}.jinja
@@ -4,7 +4,7 @@ env:
 
 # See https://github.com/OCA/odoo-community.org/issues/37#issuecomment-470686449
 parserOptions:
-  ecmaVersion: 2019
+  ecmaVersion: {% if odoo_version < 17 %}2019{% else %}2022{% endif %}
 
 overrides:
   - files:


pedrobaeza:
Please do it only for 17.0

pedrobaeza:
I think you should limit the previous template, and please also check #258

chienandalu:
> I think you should limit the previous template, and please also check #258

That issue states that the new templating is used from eslint 9.3.0, but version 8.24.0 is being used right now.

https://github.com/OCA/oca-addons-repo-template/blob/b47dbcc4e294635c539646c9931d09140e8733d7/src/.pre-commit-config.yaml.jinja#L42

pedrobaeza:
Yes, but the idea is that you also update `eslint` in this shot (if there's no incompatibility).

chienandalu:
> Let my try it locally anyway

Well, I'm finding some issues locally with how the new eslint config works in a modular way... I can't figure out how to make it work, I'm afraid.

How's that an issue in any case? Isn't it just a nice to have?

chienandalu:
Ok, tried locally and working

pedrobaeza (approved):

pedrobaeza:
@sbidoul @yajo can you check?

src/{% if 12 < odoo_version < 17 %}.eslintrc.yml{% endif %}
yajo, line 1: It seems like both eslintrc files are pretty similar. Couldn't you just name this file `{% if odoo_version > 12 %}.eslintrc.yml{% endif %}.jinja` (notice the `.jinja` suffix) and template whatever you need inside of it?

chienandalu: That was the first approach :sweat_smile: https://github.com/OCA/oca-addons-repo-template/compare/55566fb29e0ed893c6f60689e7044d9c015c75bb..5d0a238db2efed04014ed876f06a2d5dae3b1c3e

pedrobaeza: If I read correctly the previous version, it was changing the ecmaVersion for all the versions, not only v17. Maybe I was wrong.

chienandalu: oh, you both are right :smile:

yajo: I mean that you can put:

```jinja
  ecmaVersion: {% if odoo_version < 17 %}2019{% else %}2022{% endif %}
```

In a single file and avoid having 2.

chienandalu: Done :)

sbidoul (approved):
LGTM. Code review only.

pedrobaeza (approved):
Can you merge this and issue a new version? This can be applied massively to 17.0 branches. It shouldn't be too much problem, as we come from a massive update recently.

Issue #140, opened 2022-05-20
README: Add PSC's in charge


rousseldenis:
I don't know if it's possible, but this could be a help for contributors to have PSC's name and links (to odoo-community.org) in repos README.

rousseldenis:
@OCA/core-maintainers

dreispt:
Makes sense to me.
Could be an additional question for `copier`.

legalsylvain:
The information is in the OCA instance. Dont know how to recover it to générate readme...

rousseldenis:
> The information is in the OCA instance. Dont know how to recover it to générate readme...

Yes that's a point I don't have answer.

hbrunn:
isn't https://github.com/OCA/maintainer-tools/blob/master/tools/oca_projects.py where the info about PSCs/repos lives?

dreispt:
I suggested to keep it simple, and make it a copier question.

yajo:
It can be done by a controller in https://odoo-community.org, where [all that information is already stored](https://odoo-community.org/psc-teams).

For example, Copier could just put a link to https://odoo-community.org/psc-teams/find-by-repo?repo=OCA/project in the readme which just makes a 302 redirection towards https://odoo-community.org/psc-teams/project-service-28.

francesco-ooops:
Any update on this? I resorted to this workaround 😅

![image](https://user-images.githubusercontent.com/55489873/187875107-f53c66ce-cd0e-48e7-a3b0-7dfcae4ca73f.png)

legalsylvain:
> I suggested to keep it simple, and make it a copier question.

Not sure it's simple if we duplicate the information in two places. (in odoo-community + in copier answers file). There will be harder to maintain, don't you think ?

simahawk:
Good news: we are going to manage repos via github actions. I'm sure we can handle this too.
We are going to have a talk during OCA days but if you want to have a look at the prototype, here you go https://github.com/OCA-devel/repo-maintainer :wink:

pedrobaeza:
Why having a different organization for such repo?

simahawk:
> Why having a different organization for such repo?

It's temporary. We needed a sandbox to play with. We'll move what's needed to OCA when ready.

PR #261, opened 2024-05-31
fix(readme): poor man PSC indicator badge

yajo:
Provide a more ergonomic way for https://github.com/OCA/oca-addons-repo-template/issues/140#issuecomment-1233973128 to fix https://github.com/OCA/oca-addons-repo-template/issues/140.

### DIFF
--- a/src/README.md.jinja
+++ b/src/README.md.jinja
@@ -3,6 +3,7 @@
 {%- endmacro -%}
 {%- if org_slug == "OCA" %}
 [![Runboat](https://img.shields.io/badge/runboat-Try%20me-875A7B.png)](https://runboat.odoo-community.org/builds?repo={{ org_slug }}/{{ repo_slug }}&target_branch={{ odoo_version }})
+[![Find PSC team](https://img.shields.io/badge/PSC-Find%20team-875A7B.png)](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fodoo%2Dcommunity.org%2Fpsc%2Dteams%20%22{{ repo_slug | urlencode }}%22)
 {%- else %}
 <!-- /!\ Non OCA Context : Set here the badge of your runbot / runboat instance. -->
 {%- endif %}


carmenbianca (approved):
LGTM, even if it's not ideal.

Two suggestions:

- Keep an issue open somewhere to tag this as in need of improvement.
- Consider using DuckDuckGo instead of Google :)

src/README.md.jinja
```diff
@@ -3,6 +3,7 @@
 {%- endmacro -%}
 {%- if org_slug == "OCA" %}
 [![Runboat](https://img.shields.io/badge/runboat-Try%20me-875A7B.png)](https://runboat.odoo-community.org/builds?repo={{ org_slug }}/{{ repo_slug }}&target_branch={{ odoo_version }})
+[![Find PSC team](https://img.shields.io/badge/PSC-Find%20team-875A7B.png)](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fodoo%2Dcommunity.org%2Fpsc%2Dteams%20%22{{ repo_slug | urlencode }}%22)
```
legalsylvain, line 6:
```suggestion
[![Find PSC team](https://img.shields.io/badge/PSC-Find%20team-875A7B.png)](https://duckduckgo.com/?q=site%3Ahttps%3A%2F%2Fodoo-community.org%2Fpsc-teams+%22{{ repo_slug | urlencode }}%22)
```

this should work @carmenbianca

sbidoul:
We should rather link to https://oca.github.io/repo-maintainer-conf/, because https://odoo-community.org/psc-teams.

With some tweaks to https://github.com/OCA/repo-maintainer-conf/blob/master/.github/workflows/static.yml and the script it invokes, we should even be able to have a rich man solution.

@simahawk @etobella are we ready to remove https://odoo-community.org/psc-teams or replace it with redirects to ttps://oca.github.io/repo-maintainer-conf ?

src/README.md.jinja
yajo: Sadly, not:

image

yajo:
> We should rather link to https://oca.github.io/repo-maintainer-conf/, because https://odoo-community.org/psc-teams is unmaintained.

Sadly, we can't either:
image

yajo:
> With some tweaks to https://github.com/OCA/repo-maintainer-conf/blob/master/.github/workflows/static.yml and the script it invokes, we should even be able to have a rich man solution.

Do you think it's worth waiting for this to happen? Or could we merge this workaround and just update the badge url once the rich man solution lands?

legalsylvain:
Do you think it's worth waiting for this to happen?

As it is changing the template of OCA modules, better to not push a temporary solution IMO that will be replaced by another in a few time. Matter of taste, though.

PR #263, opened 2024-06-27
Add Debian / Redhat / MacOS packages to gitignore

gurneyalex:


gurneyalex:
Maybe we need other package managers in there too (I don't know what you get on MacOS for instance, and there are probably OCA contributors using that OS)

pedrobaeza (approved):

pedrobaeza:
`.dmg` for MacOS, so OK to include it as well.

carmenbianca (approved):

pedrobaeza:
@sbidoul are you going to do a massive deployment of the new copier template?

PR #271, opened 2024-09-26
[IMP] Set OCA_ENABLE_CHECKLOG_ODOO by default starting with Odoo 18.0

PabloEForgeFlow:


PabloEForgeFlow:
@sbidoul I added a question for the setting, it does seem more appropriate.

Regarding version 18.0, now that both the odoo and ocb branches are up, is there something keeping us from adding support here in the template, besides the [oca-ci images](https://github.com/OCA/oca-ci/blob/master/.github/workflows/ci.yaml#L23) not existing yet? Let me know if I can help somehow.

copier.yml
```diff
@@ -157,6 +157,11 @@ include_wkhtmltopdf:
     Do you need to install wkhtmltopdf? Usually only needed if you're going to test PDF
     report generation.
 
+enable_checklog_odoo:
+  type: bool
+  default: "{% if odoo_version < 18.0 -%}no{% else %}yes{% endif %}"
+  help: Do you want warnings to make tests fail?
```
sbidoul, line 163:
```suggestion
  help: Do you want warnings to make tests fail on warnings?
```

sbidoul:
Thanks! Just a minor remark.

> Regarding version 18.0, now that both the odoo and ocb branches are up, is there something keeping us from adding support here in the template, besides the [oca-ci images](https://github.com/OCA/oca-ci/blob/master/.github/workflows/ci.yaml#L23) not existing yet? Let me know if I can help somehow.

We can move forward with that yes. The part that requires most effort is modernizing the linter versions and configs, including pylint-odoo and friends.

copier.yml
pedrobaeza: Shouldn't it be?

```suggestion
  help: Do you want to make tests fail on warnings?
```

PabloEForgeFlow: Fixed

sbidoul:
@PabloEForgeFlow this does not always work, because the you added the new env variable in an already conditional section. See https://github.com/OCA/server-env/blob/18.0/.github/workflows/test.yml

PR #272, opened 2024-09-26
Add version 18

SirAionTech:


src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -12,6 +12,7 @@ on:
 {%
 set IMAGES = {
   "odoo": {
+    18.0: "ghcr.io/oca/oca-ci/py3.10-odoo18.0:latest",
```
amh-mw, line 15: python 3.10 here doesn't match python-version 3.12 above?

pedrobaeza:
The question here is also to bump the linters versions up to the recommended/compatible ones.

.github/workflows/test.yml
```diff
@@ -35,6 +35,9 @@ jobs:
           - odoo-version: 17.0
             python-version: "3.10"
             machine: ubuntu-22.04
+          - odoo-version: 18.0
+            python-version: "3.12"
```
sbidoul, line 39: We'll test with python 3.10, since this is the minimum supported version, per
https://github.com/odoo/odoo/blob/9e35989436b807d5db738116cd3982dcb4efe81c/setup.py#L66

To be revisited/confirmed when Odoo release their official Docker image.

sbidoul: Oh, but this is to test the template itself, so ok.

src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
sbidoul: 3.10 is correct, per
https://github.com/odoo/odoo/blob/9e35989436b807d5db738116cd3982dcb4efe81c/setup.py#L66

SirAionTech:
The `pyyaml` error in https://github.com/OCA/oca-addons-repo-template/actions/runs/11054048607/job/30709905165?pr=272#step:7:90 looks like a known issue https://github.com/yaml/pyyaml/issues/601

sbidoul:
> The `pyyaml` error in https://github.com/OCA/oca-addons-repo-template/actions/runs/11054048607/job/30709905165?pr=272#step:7:90 looks like the known issue [yaml/pyyaml#601](https://github.com/yaml/pyyaml/issues/601)

I continue to think we should not use poetry nor pin dependencies to test this repo. cc/ @yajo

SirAionTech:
> The question here is also to bump the linters versions up to the recommended/compatible ones.

There you go: https://github.com/OCA/oca-addons-repo-template/pull/272/commits/5cf221cdf985358d0c2f800281d5775f21ab3e44

`prettier` could also be updated but:

- https://github.com/pre-commit/mirrors-prettier is archived
- `prettier 3.0.3` with the `plugin-xml` is affected by https://github.com/prettier/prettier/issues/15696
- https://prettier.io/docs/en/precommit.html gives some instructions but to me it looks like they have nothing to do with https://github.com/pre-commit/pre-commit

I'd remove that for another alternative but don't know any, so I left it as it was

SirAionTech:
@sbidoul using Python 3.9 fixes https://github.com/OCA/oca-addons-repo-template/actions/runs/11054931643/job/30712926871#step:8:309
> AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?

but breaks builds on Ubuntu 20.04 so I have removed them, CI is still requiring them because of some hidden setting I think (that should now include 17 and 18), let me know if I have to do anything

src/.pre-commit-config.yaml.jinja
```diff
@@ -238,7 +238,7 @@ repos:
         additional_dependencies: ["flake8-bugbear=={{ repo_rev.flake8_bugbear }}"]
   {%- else %}
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.3
+    rev: v0.6.8
```
sbidoul, line 241: I think we should have a variable for this ruff version, so we can control which one we use with each Odoo version.

src/.pre-commit-config.yaml.jinja
```diff
@@ -39,17 +39,17 @@
...
+  {%- set repo_rev.maintainer_tools = "bf9ecb9938b6a5deca0ff3d870fbd3f33341fded" %}
   {%- set repo_rev.nodejs = "16.17.0" %}
-  {%- set repo_rev.odoo_pre_commit_hooks = "v0.0.25" %}
-  {%- set repo_rev.pre_commit_hooks = "v4.3.0" %}
+  {%- set repo_rev.odoo_pre_commit_hooks = "v0.0.33" %}
+  {%- set repo_rev.pre_commit_hooks = "v4.6.0" %}
   {%- set repo_rev.prettier = "2.7.1" %}
   {%- set repo_rev.prettier_xml = "2.2.0" %}
-  {%- set repo_rev.pylint_odoo = "v8.0.19" %}
+  {%- set repo_rev.pylint_odoo = "v9.1.2" %}
```
sbidoul, line 52: Is it safe to upgrade all linters for 16 and 17 too? What is the risk of making existing repos red due to new checks?

SirAionTech: > Is it safe to upgrade all linters for 16 and 17 too?

Is it safe to keep using the old ones? For how long should we keep using them?
I'm for using the new and shiny things, however I'll change it if you think it's better.

> What is the risk of making existing repos red due to new checks?

I'll try the update in a repo to see how much it should change to satisfy the new requirements.

I could have left everything < 18 crystallized as it was but maybe they should be updated too from time to time, don't they?

src/.pre-commit-config.yaml.jinja
SirAionTech: Is `ruff` particular? Otherwise I'd fall back to https://github.com/OCA/oca-addons-repo-template/pull/272#discussion_r1778137863:

> I could have left everything < 18 crystallized as it was but maybe they should be updated too from time to time, don't they?

src/.pre-commit-config.yaml.jinja
sbidoul: > I could have left everything < 18 crystallized as it was but maybe they should be updated too from time to time, don't they?

I don't think we should update for previous branches unless necessary. Otherwise tiny changes in linters will make existing branches red, and this is really annoying.

So I would add a new section for 18, with the latest linters.

yajo: Old branches won't be affected unless manually updated. And in that case, pre-commit will update things and maybe ask for new changes, but that can be considered as part of that same manual work. However, it's true that if there must be a mass update of non-pre-commit-related things, that can be problematic. Oftentimes you're the one who does that, @sbidoul, so as you wish.

yajo (approved):

sbidoul:
Let's try this. Thanks a lot @SirAionTech !

PR #274, opened 2024-10-01
[FIX] oca-check-po: don't check po-pretty-format

rousseldenis:
Don't enforce po pretty format that raises if a msgid has the same value as the msgstr. As weblate has those values, it will lead to problems.

Moreover, in some languages, some translations can have the same value as the original language.

@moylop260 @sbidoul 

Related to : https://github.com/OCA/server-env/pull/204

Needed for v18.

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -159,6 +159,8 @@ repos:
     hooks:
       - id: oca-checks-odoo-module
       - id: oca-checks-po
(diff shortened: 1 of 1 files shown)


sbidoul:
The main problem is the error if the translated string is equal to the source string.
To me it is important to not fail in that case, because it happens for legitimate reason, and if we keep the translation empty, weblate will consider that the string is not translated, which is not ok for translators.

PR #275, opened 2024-10-01
[FIX] Move enable_checklog_odoo check outside of nested conditional

PabloEForgeFlow:


PabloEForgeFlow:
@sbidoul as you mentioned on https://github.com/OCA/oca-addons-repo-template/pull/271, I made `enable_checklog_odoo` depend on the outer condition.

I fixed that and took the liberty of doing a little refactor to avoid cluttering the outer conditional and improve the clarity of the template, I don't know if it did. Let me know what you think.

sbidoul:
Sounds good.

The jinja2 docs talks about a `do` statement that looks cleaner, but I don't know what it takes to enable that. It looks like it's a "extension".

So let's move ahead with this version.

PabloEForgeFlow:
> The jinja2 docs talks about a `do` statement that looks cleaner, but I don't know what it takes to enable that. It looks like it's a "extension".

It appears that copier allows [enabling extensions](https://copier.readthedocs.io/en/stable/configuring/#jinja_extensions) in the `copier.yml` file. If there's no drawback to enabling it I guess we could do that.

Issue #264, opened 2024-06-27
pre-commit/mirrors-prettier has been archived


sbidoul:
The https://github.com/pre-commit/mirrors-prettier repo has been archived without much explanation.

There is some discussion in https://github.com/prettier/prettier/issues/15742

We may need to look for an alternative.

PabloEForgeFlow:
I was looking at some of the proposed solutions in the discussion, and using [local hooks](https://pre-commit.com/#repository-local-hooks) (https://github.com/prettier/prettier/issues/15742#issuecomment-2180420244) seems quite reasonable to me. We would need to pull in all the keys from [.pre-commit-hooks.yaml](https://github.com/pre-commit/mirrors-prettier/blob/main/.pre-commit-hooks.yaml), but I think it's simple enough that it shouldn't be a problem.

We could perhaps do the same for `eslint` and remove the dependency on the repo, in case something similar happens to it.

sbidoul:
@PabloEForgeFlow that sounds worth trying, yes.

PabloEForgeFlow:
Good to hear, I'll get something working and do a PR.

We will probably want to update the version for both of them in v18 right? They've both had new major versions with breaking changes (`prettier` in [v3.0.0](https://prettier.io/blog/2023/07/05/3.0.0.html) and `eslint` in [v9.0.0](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/)), so I'm not sure if we should only update to the latest minor version.

sbidoul:
Yes let's use the latest versions in 18. If we don't wait too much, we can still push it to the 18.0 branches. The more we wait, the more we risk making branches red and create trouble for maintainers, though.

PabloEForgeFlow:
@sbidoul Hey Stéphane, could you take a look at https://github.com/OCA/oca-addons-repo-template/pull/278?

sbidoul:
Yes yes, it's on my radar :sweat:

PabloEForgeFlow:
> Yes yes, it's on my radar 😓

Sorry, I didn't mean to pressure you. I thought the notification might have gotten lost among the many things that are going on. :bow:

sbidoul:
No worries :) I made a couple of quick comments, and I'll try to test this week still.

Issue #224, opened 2023-11-04
Improve xml prettyfier


sbidoul:
prettier's xml pluggin now supports a much better XML autoformatting mode. See https://github.com/prettier/plugin-xml/issues/138#issuecomment-724547415.

Below are my findings on how to make it work.

You need a `.prettierrc.cjs` instead of `.prettierrc.yml`, so prettier can load the plugin even when run from pre-commit:

```javascript
/** @type {import('prettier').Config} */
const config = {
  plugins: [require.resolve("@prettier/plugin-xml")],
  bracketSpacing: false,
  printWidth: 88,
  proseWrap: "always",
  semi: true,
  trailingComma: "es5",
  xmlWhitespaceSensitivity: "preserve",
};

module.exports = config;
```

And the pre-commit config needs to be adapted like so:

```yaml
  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: v3.0.3
    hooks:
      - id: prettier
        name: prettier + plugin-xml
        additional_dependencies:
          - "prettier@3.0.3"
          - "@prettier/plugin-xml@3.2.2"
```

yajo:
> You need a `.prettierrc.cjs` instead of `.prettierrc.yml`, so prettier can load the plugin even when run from pre-commit

Prettier was always able to load the XML plugin until now. Did something relevant change in that regard?

sbidoul:
Apparently prettier 3 changed module discovery in a way that is incompatible with pr-commit. It's a rabbit hole and this approach is the best I could find

PR #278, opened 2024-10-08
[IMP] pre-commmit: Run prettier and eslint as local hooks and update versions in v18

PabloEForgeFlow:
Should fix https://github.com/OCA/oca-addons-repo-template/issues/264 and https://github.com/OCA/oca-addons-repo-template/issues/224.


PabloEForgeFlow:
@sbidoul I made the `prettier` and `eslint` hooks local and updated versions at the same time. Perhaps they should be two separate PRs?

Also, pre-commit was [failing](https://github.com/OCA/oca-addons-repo-template/actions/runs/11249365414/job/31276135881), due to the `.prettierrc.yml` symlink being broken, so I updated it to the new `'{% if 12 < odoo_version < 18 %}.prettierrc.yml{% endif %}'`, but we'll probably want to use the v18 file once we're sure it's working.

sbidoul:
Is it not necessary to convert eslintrc.yml to cjs, and to the new eslint 9 config format for v18?

We could also upgrade node for v18.

BTW, I wonder why GitHub is still trying to run the tests for 11 and 12.

PabloEForgeFlow:
> Is it not necessary to convert eslintrc.yml to cjs, and to the new eslint 9 config format for v18?

Will do that, I thought it was something limited to prettier, but it does appear to have [changed as well](https://eslint.org/docs/latest/use/configure/configuration-files). Guess that's the new way of doing configuration files in the javascript world.

> We could also upgrade node for v18.

I will look into it, see what version makes most sense.

> BTW, I wonder why GitHub is still trying to run the tests for 11 and 12.

Yeah that's weird, maybe something's still cached on GitHub's side?

PabloEForgeFlow:
I've migrated the eslint configuration to new format following both [eslint](https://eslint.org/docs/latest/use/configure/migration-guide) and [jsdoc](https://github.com/gajus/eslint-plugin-jsdoc/wiki/Comparison-with-deprecated-JSdoc-related-ESLint-rules) guides. Most parts have been migrated using the [configuration migrator](https://www.npmjs.com/package/@eslint/migrate-config), but some, such as jsdoc settings, were done by hand. I also set `ecmaVersion` to 2024, but I'm not sure if we would rather stay on 2022.

I also update the node to version 22, which is the latest TLS release, and fixed some issues with prettier.

sbidoul:
Hi @PabloEForgeFlow I did a test by applying the template from your PR onto https://github.com/OCA/storage/pull/399

I noticed two things:

1. In 18+, we need to remove `.prettierrc.yml` otherwise it takes precedence over the cjs config. We also need to remove the old `.eslintrc.yml`. This can be done in `copier.yml`, `_tasks`.
2. I think we need to exclude `^eslint.config.cjs|^prettier.config.cjs`, otherwise eslint is trying to check them and is not happy with what it sees.

PabloEForgeFlow:
Makes sense, I added a line to `_tasks` removing them and put both in the global pre-commit exclude.

copier.yml
```diff
@@ -5,6 +5,7 @@ _min_copier_version: "9"
 
 _tasks:
   - rm -f {% if ci != "Travis" %}.travis.yml{% endif %} .t2d.yml
+  - rm -f .prettierrc.yml .eslintrc.yml
```
sbidoul, line 8: @PabloEForgeFlow these should be removed only for odoo_version >= 18, right? Can you adjust that?

sbidoul: Done in 6f0c78888e983606697d4d4c8cf52bf4c3eec471

PabloEForgeFlow: I though they would be recreated, but yeah it's probably more clear this way.

PR #279, opened 2024-10-11
[FIX] ruff: ignore printf-string-formatting of rule up031

sanderlienaerts:
When not using Ruff, pre-commit prevents pyupgrade from replacing printf with str.format() (https://github.com/OCA/oca-addons-repo-template/pull/49). 

Using Ruff this rule is not being ignored. This commit modifies the Ruff configuration adding an ignore rule for `UP031`.

### DIFF
--- a/src/{% if use_ruff %}.ruff.toml{% endif%}.jinja
+++ b/src/{% if use_ruff %}.ruff.toml{% endif%}.jinja
@@ -18,6 +18,7 @@ extend-select = [
     "{{ rule }}",
     {%- endfor %}
 ]
+ignore = ["UP031"]
 exclude = ["setup/*"]
 
 [format]


pedrobaeza:
In which cases ruff is not used?

sanderlienaerts:
> In which cases ruff is not used?

Ruff is not being used in version < 14 or when `use_ruff` is set to `False` in Copier by the user.

sbidoul:
I think we should not disable UP031.

If my memory is correct, using `.format` is only a risk for translatable strings, and that is caught by pylint Odoo.
And in Odoo the best pattern for translatable strings is `_("aaa %s bbb %s", 1, 2)`, which is not handled by UP031.

In other cases, that rule is useful, because f-strings are generally more readable that printf style.

sanderlienaerts:
> If my memory is correct, using `.format` is only a risk for translatable strings, and that is caught by pylint Odoo. And in Odoo the best pattern for translatable strings is `_("aaa %s bbb %s", 1, 2)`, which is not handled by UP031.

I have just tested this on the pre-commit config, confirming that this is the current working behaviour. So since formatting conventions for translatable strings are upheld, I will close the PR. Thanks for your feedback.

PR #280, opened 2024-10-11
[FIX] pre-commit: update pylint odoo to enable valid-odoo-versions key for v18

sanderlienaerts:
Without updating the linter and using `valid-odoo-versions=18.0`, pylint_odoo (mandatory check) passes green but without checking any rules.

pylint-odoo unfortunately doesn't notify the user of 18.0 not being supported in v9.1.2. Support was added in v9.1.3 (https://github.com/OCA/pylint-odoo/compare/v9.1.2...v9.1.3) in PR https://github.com/OCA/pylint-odoo/pull/497

Example with `valid-odoo-versions=18.0` and `pylint-odoo==v1.9.2`:
Screenshot 2024-10-11 at 13 54 37

Working example with `valid-odoo-versions=18.0` and `pylint-odoo==v1.9.3`:
Screenshot 2024-10-11 at 13 53 57

Builds on: https://github.com/OCA/oca-addons-repo-template/pull/272

pylint_odoo's version was also being overridden. This commit fixes that by adding a new v17 section.

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -38,7 +38,7 @@
   {%- set repo_rev.pyupgrade = "v2.29.0" %}
   {%- set repo_rev.ruff = "v0.1.3" %}
   {%- set repo_rev.setuptools_odoo = "3.1.8" %}
-{%- elif odoo_version < 18 %}
+{%- elif odoo_version < 17 %}
   {%- set repo_rev.autoflake = "v1.6.1" %}
   {%- set repo_rev.black = "22.8.0" %}
   {%- set repo_rev.eslint = "8.24.0" %}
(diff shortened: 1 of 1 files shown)


src/.pre-commit-config.yaml.jinja
```diff
@@ -68,16 +68,12 @@
...
   {%- set repo_rev.prettier_xml = "2.2.0" %}
-  {%- set repo_rev.pylint_odoo = "v9.1.2" %}
+  {%- set repo_rev.pylint_odoo = "v9.1.3" %}
   {%- set repo_rev.pyupgrade = "v2.38.2" %}
   {%- set repo_rev.ruff = "v0.6.8" %}
   {%- set repo_rev.setuptools_odoo = "3.1.8" %}
 {%- endif %}
 
-{%- if odoo_version > 16 %}
-  {%- set repo_rev.pylint_odoo = "v9.0.4" %}
```
rven, line 78: By removing this v17 will have pylint_odoo = "v8.0.19" instead of "v9.0.4"

rven: {%- if odoo_version == 17 %}
  {%- set repo_rev.pylint_odoo = "v9.0.4" %}
{%- endif %}

sanderlienaerts: Ah right, miss interpreted the conditions. So if `v9.0.4` only applies to 17, we shouldn't remove the condition but instead modify to check for 17 only. In the previous state, the `v9.1.2` for v17 and v18 were changed back to `v9.0.4`.

sbidoul: Tihs is indeed easy to miss. I'd prefer adding a whole new section for v17, even if all the versions are identical to 16. So it is systematic, and easier to understand.

sbidoul:
> pylint-odoo unfortunately doesn't notify the user of 18.0 not being supported

TIL. I'm adding a new checkbox in https://github.com/OCA/maintainer-tools/issues/628

sbidoul:
Thanks! Can you use the same versions in the 16 and 17 sections (at first sight it looks like you cloned the 18 section for 17) ?

rven:
> Thanks! Can you use the same versions in the 16 and 17 sections (at first sight it looks like you cloned the 18 section for 17) ?

I think it's correct what has been done here.

sanderlienaerts:
Changes are incorporated from `origin/master`

PR #281, opened 2024-10-16
[IMP] Add new workflow to remove merged branches.

carlosdauden:
The idea is that this action is executed in the repository fork to avoid permission problems, but it is included in the main repo to facilitate its propagation.

I am not very familiar with how these tools work, so this should be taken more as an idea to be developed than as a PR.

@Tecnativa
ping @josep-tecnativa

### DIFF
--- a/.github/workflows/clean-merged-branch.yml
+++ b/.github/workflows/clean-merged-branch.yml
@@ -1,0 +1,17 @@
+name: Delete merged branches in fork
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  delete_branch:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true
+    steps:
+      - name: Delete merged branch from fork
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git push origin --delete ${{ github.event.pull_request.head.ref }}


pedrobaeza (approved):
This is very interesting for keeping the environment cleaned

josep-tecnativa:
This looks good to me.

legalsylvain:
I'm not sure to understand which branch will be deleted. sorry.

Let's take an exemple. If Tecnativa team makes a PR on OCA/web against OCA:16.0, named tecnativa:16.0-IMP-something, Does it mean the branch tecnativa:16.0-IMP-something will be deleted once the PR is merged in OCA:16.0 ?

thanks for the precision.

legalsylvain:
> That is exactly the idea.

Thanks ! 

Well, I'm not totally sure of this script. (specially on access right : The event is raised "on" the OCA repo, when a PR is merged, but the script ask to delete a branch of the "tecnativa" organisation.) 

Could you test it in a custom place, before we deploy this solution in all the OCA repo ? 
(for exemple add the code in tecnativa:16.0 and then create a branch carlosdauden:16.0-TEST and make a PR ? )

Thanks !

pedrobaeza:
@legalsylvain that's not the idea. This should be put on the OCA repos, so that when you fork the branch in your own repository, it acts like explained. Obviously, if someone in OCA does an extra branch, propose a pull request, and gets merged, the branch in OCA will be removed as well.

Issue #175, opened 2022-11-28
[RFC] Replace pre-commit with precommix or pre-commit-hooks.nix


yajo:
**Is your feature request related to a problem?**
Here at OCA we use [pre-commit](https://pre-commit.com) extensively to make it more comfortable to share the same style of programming. It also bundles a lot of tools that help the coding and reviewing process.

However, pre-commit itself is divided into 3 parts:

1. Install the hooks.
2. Find the hooks.
3. Run the hooks.

While it's perfect for running the hooks, and pretty OK for finding them, it's not perfect for installing them. How can that be, if we pin the hook tags? Because everything that's below the hook (python version, or node version, or stdlib version, OS version, etc.) is not pinned.

The obvious outcome of this problem is that every now and then, some hook dependency starts failing, and then the CI and dev experiencie starts melting across all OCA repos. It becomes a maintenance burden with ease. Here I've collected some examples (related but not exclusive to OCA):

1. https://github.com/OCA/oca-addons-repo-template/pull/41
2. https://github.com/OCA/oca-addons-repo-template/issues/133
3. https://github.com/OCA/oca-addons-repo-template/pull/143
4. https://github.com/OCA/mis-builder/pull/476#issuecomment-1313663536
5. https://github.com/OCA/oca-addons-repo-template/issues/174
6. https://github.com/OCA/oca-addons-repo-template/pull/182
7. https://github.com/OCA/l10n-spain/pull/2783#issuecomment-1408349114
8. https://github.com/OCA/oca-addons-repo-template/pull/183
9. https://github.com/OCA/oca-addons-repo-template/pull/184
10. https://github.com/sbidoul/manifestoo/pull/53#issuecomment-1415764949
11. https://github.com/acsone/click-odoo-contrib/pull/126#issuecomment-1411805318
12. https://github.com/OCA/oca-addons-repo-template/issues/80
13. https://github.com/Tecnativa/doodba-copier-template/pull/370
14. https://github.com/OCA/oca-addons-repo-template/pull/188
15. https://github.com/OCA/oca-addons-repo-template/issues/193
16. https://github.com/OCA/oca-addons-repo-template/pull/194
17. https://github.com/OCA/oca-addons-repo-template/pull/196
18. https://github.com/OCA/oca-addons-repo-template/pull/199
19. https://github.com/OCA/oca-addons-repo-template/issues/200
20. https://github.com/OCA/oca-addons-repo-template/pull/198#issuecomment-1645652163
[...]

PR #282, opened 2024-10-23
[IMP] Use specific python version for a set of hooks

rousseldenis:
Pin the python version per Odoo version.

e.g.: as each local environment can be different, force the python version to avoid problems with incompatible hooks versions

@sbidoul

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -3,6 +3,7 @@
 
 {#- '<15' only affects to 14.0 #}
 {%- if odoo_version < 15 %}
+  {%- set repo_rev.python_version = "python3.8" %}
   {%- set repo_rev.autoflake = "v1.4" %}
   {%- set repo_rev.black = "22.3.0" %}
   {%- set repo_rev.eslint = "7.8.1" %}
@@ -21,6 +22,7 @@
   {%- set repo_rev.ruff = "v0.1.3" %}
   {%- set repo_rev.setuptools_odoo = "3.1.8" %}
 {%- elif odoo_version < 16 %}
+  {%- set repo_rev.python_version = "python3.8" %}
   {%- set repo_rev.autoflake = "v1.4" %}
   {%- set repo_rev.black = "22.3.0" %}
   {%- set repo_rev.eslint = "7.32.0" %}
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is a bit problematic, as each Linux distribution may come with a different Python version pre-installed. What if you have Python 3.9 and is compatible, and you force 3.8? Do you have to install (and may get SO conflicts) to get Python 3.8 to run it? I understand that the goal is to avoid the problem if pre-commit for that version doesn't work for example for Python 3.10, but the ideal things is to have a supported Python versions interval. What I don't know is how to perform that.

carmenbianca:
I think most experienced Odoo devs know how to simultaneously have multiple versions of Python installed on their machine. This shouldn't be a concern for them.

The tricky bit is that lesser-experienced devs may not, and this creates a _huge_ barrier to entry. For me, this barrier to entry makes this PR a non-starter, even though I've often wanted to make the same PR.

The way I am currently circumventing this problem is by installing pre-commit in my Odoo virtual environment, which uses a compatible version of Python. This isn't perfect, because when you switch a repo's branch to a different Odoo version, you may be using the wrong virtual environment's pre-commit.

@yajo may want to add this to #175 again.

yajo:
One example: [ruff only supports target version >= 3.7](https://docs.astral.sh/ruff/settings/#target-version). However, [Odoo 13 ran on python 3.6](https://github.com/Tecnativa/doodba/blob/0c5b619ecf3c8275863bda219d58abd221a77b72/13.0.Dockerfile#L1). So we cannot use ruff for formatting an Odoo 13 codebase.

Just an example; Ruff didn't exist back then, and i's done in Rust. But anyway I mean that this is something that sooner or later had to happen... At some point, the old pinned pre-commit deps won't be able to run on more modern python interpreters. Another option is to update them, so they do run on them. But even these deps usually deprecate support for deprecated python interpreter versions.

Yes, #175 is the definitive solution, so I won't block or push any other solution here 😁

rousseldenis:
> This is a bit problematic, as each Linux distribution may come with a different Python version pre-installed. What if you have Python 3.9 and is compatible, and you force 3.8? Do you have to install (and may get SO conflicts) to get Python 3.8 to run it? I understand that the goal is to avoid the problem if pre-commit for that version doesn't work for example for Python 3.10, but the ideal things is to have a supported Python versions interval. What I don't know is how to perform that.

@pedrobaeza The problem is that a version of a hook can work for a set of python versions and not another one. So, if you have python3.12 (my case) installed, you should use python3.12 for pre-commit install. In this case, flake8 breaks and does not work.

Maybe exploring ruff should be great as it would require just one hook. I suppose the PR's curve on v13 branches will decrease and we can live with pre-commit on those repos.

pedrobaeza:
Well, if we can put some instructions in place for run it, similar to what I did with https://github.com/OCA/maintainer-tools/wiki/Install-pre-commit, it's OK for me.

PR #288, opened 2024-12-05
[IMP] consider unsafe fixes for UP008 as safe

trisdoan:
### Context
- Take over this improvement: https://github.com/OCA/odoo-module-migrator/pull/77
- Instead of implementing in odoo-module-migrator (not always installed), do it directly with Ruff pre-commit hook
### This change
- As [this rule](https://docs.astral.sh/ruff/rules/super-call-with-parameters/) is taken into account in Ruff, simply applying its fix when with [extend-safe-fixes](https://docs.astral.sh/ruff/settings/#lint_extend-safe-fixes)

### DIFF
--- a/src/{% if use_ruff %}.ruff.toml{% endif%}.jinja
+++ b/src/{% if use_ruff %}.ruff.toml{% endif%}.jinja
@@ -18,6 +18,7 @@ extend-select = [
     "{{ rule }}",
     {%- endfor %}
 ]
+extend-safe-fixes = ["UP008"]
 exclude = ["setup/*"]
 
 [format]


sbidoul (approved):

bosd (approved):

sbidoul:
Do we know if ruff UP008 preserves exoctic uses of super that we have in Odoo? Things like `super(Class, self.filtered(...))._method()`.

trisdoan:
> Do we know if ruff UP008 preserves exoctic uses of super that we have in Odoo? Things like `super(Class, self.filtered(...))._method()`.

Hello @sbidoul, yes it does
For example, it does not impact this case

```
return super(
            QueueJob,
            self.with_context(mail_create_nolog=True, mail_create_nosubscribe=True),
        ).create(vals_list)
```

sbidoul:
Great. Let me know when you have updated the PR to remove the version test.

PR #289, opened 2024-12-24
[IMP] add pylint checkers for translation

trisdoan:
### Note
- Enabling these checks as mandatory will need some fixing work on most repos
- So if we agree that these checks are welcome, it might be better not to enable them in mass, but repo by repo

### This change
- Since https://github.com/OCA/pylint-odoo/pull/423/commits/2db6690791efa93ce06f81be40c3339462015b50, some checkers for translation were added, by emitting translation-* instead of logging-

[translation-format-interpolation](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-format-interpolation.html)
[translation-format-truncated](https://pylint.pycqa.org/en/latest/user_guide/messages/error/logging-format-truncated.html)
[translation-fstring-interpolation](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html)
[translation-not-lazy](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-not-lazy.html)
[translation-too-few-args](https://pylint.pycqa.org/en/latest/user_guide/messages/error/logging-too-few-args.html)
[translation-too-many-args](https://pylint.readthedocs.io/en/latest/user_guide/messages/error/logging-too-many-args.html)
[translation-unsupported-format](https://pylint.pycqa.org/en/latest/user_guide/messages/error/logging-unsupported-format.html)


src/.pylintrc.jinja
```diff
@@ -27,5 +27,8 @@
     redefined-builtin,
     too-complex,
     unnecessary-utf8-coding-comment
+{%- if odoo_version >= 18 %},
+    prefer-env-translation
+{%- endif %}
```
yajo, line 32: It seems the rest of checks are alphabetically-sorted. Could you do the same here please?

yajo (approved):
LGTM. A couple of sorting suggestions, nothing required or very critical.

src/.pylintrc-mandatory.jinja
```diff
@@ -69,6 +69,15 @@ enable=anomalous-backslash-in-string,
...
     xml-syntax-error
+{%- if odoo_version >= 14 %},
+    translation-format-interpolation,
+    translation-format-truncated,
+    translation-fstring-interpolation,
+    translation-not-lazy,
+    translation-too-few-args,
+    translation-too-many-args,
+    translation-unsupported-format
+{%- endif %}
```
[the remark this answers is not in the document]
trisdoan: Hi @yajo, thanks for reviewing, I updated the code

sbidoul (changes requested):
That will break too much existing stuff. Please make it a copier question (default = disabled) so PSC can enable on demand.

copier.yml
```diff
@@ -205,3 +205,8 @@ github_ci_extra_env:
 convert_readme_fragments_to_markdown:
   default: "{% if odoo_version < 17.0 -%}no{% else %}yes{% endif %}"
   type: bool
+
+enable_pylint_check_translation:
+  type: bool
+  default: false
```
sbidoul, line 211:
```suggestion
  default: "{% if odoo_version < 19.0 -%}no{% else %}yes{% endif %}"
```

src/.pylintrc-mandatory.jinja
```diff
@@ -65,6 +65,15 @@ enable=anomalous-backslash-in-string,
...
     translation-field,
     translation-required,
+{%- if enable_pylint_check_translation %},
+    translation-format-interpolation,
+    translation-format-truncated,
+    translation-fstring-interpolation,
+    translation-not-lazy,
+    translation-too-few-args,
+    translation-too-many-args,
+    translation-unsupported-format
```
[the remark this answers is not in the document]
trisdoan: yes I agree, blocking is not necessary

copier.yml
```diff
@@ -205,3 +205,8 @@ github_ci_extra_env:
 convert_readme_fragments_to_markdown:
   default: "{% if odoo_version < 17.0 -%}no{% else %}yes{% endif %}"
   type: bool
+
+enable_pylint_check_translation:
+  type: bool
+  default: "{% if odoo_version < 19.0 -%}no{% else %}yes{% endif %}"
+  help: Enable pylint checks for translation?
```
yajo, line 212: Since you're adding the checks to the non-required list, I don't think we need an extra option for this. They can be just enabled always IMHO.

yajo (changes requested):

src/.pylintrc.jinja
```diff
@@ -24,8 +24,20 @@
     no-utf8-coding-comment,
     odoo-addons-relative-import,
     old-api7-method-defined,
+{%- if odoo_version >= 18 %}
+    prefer-env-translation,
+{%- endif %}
     redefined-builtin,
     too-complex,
     unnecessary-utf8-coding-comment
```
etobella, line 32: missing coma here,

might be better to put all of it befor redefined-builtin, isn't it?

PR #290, opened 2025-01-03
Push .pot changes without GitHub secret

sbidoul:
This is a bit more verbose, but avoid the maintenance of a secret token.

Leaving as draft for now.


src/.github/workflows/{% if ci == 'GitHub' %}test.yml{% endif %}.jinja
```diff
@@ -167,6 +167,34 @@ jobs:
       {%- endif %}
       {% raw -%}
       - name: Update .pot files
-        run: oca_export_and_push_pot https://x-access-token:${{ secrets.GIT_PUSH_TOKEN }}@github.com/${{ github.repository }}
-      {%- endraw %}
+        run: |
+          oca_export_and_commit_pot
+          git format-patch --stdout --keep-subject @{u}..@ > po.patch
```
sbidoul, line 172: The `click-odoo-makepot` command creates a distinct commit for each addon where pot files have been updated. So by using a patch, we export this list of commits to pass it to the job that will push them to the repo.

PR #298, opened 2025-03-06
Pre-commit config Various Versionbumps for V18

bosd:
Update .pre-commit-config.yaml.jinja


sbidoul:
Unfortunately we can't do this, as a mass update of the template to all repos would make many branches red due to new checks or lint rules enabled by default.

What we could do instead is add copier questions to set the version for the main linters: ruff, pylint-odoo, oca-pre-commit-hooks, so each repo can evolve at its own pace and enable new versions when they want to.

PR #299, opened 2025-03-25
[IMP] eslint config: include browser globals such as 'document'

StefanRijnhart:
Fixes
```
  22:23  error  'document' is not defined  no-undef
```

in code like

```
const image = document.createElement("img");
```

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -79,6 +79,7 @@
   {%- set repo_rev.eslint_jsdoc = "50.3.1" %}
   {%- set repo_rev.flake8 = "3.9.2" %}
   {%- set repo_rev.flake8_bugbear = "21.9.2" %}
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@chienandalu is the use of that global sane?

StefanRijnhart:
@pedrobaeza FYI it is the equivalent of `env: browser` that we had previously (https://github.com/eslint/eslint/discussions/16975)
https://github.com/OCA/oca-addons-repo-template/blob/master/src/{%25 if 12 < odoo_version < 18 %25}.eslintrc.yml{%25 endif %25}.jinja#L2

chienandalu (approved):
Thanks for the explanations @StefanRijnhart 

I wasn't aware of that limitation in the linter.

Also, looks like there are some dated globals like the ones for underscore or moment.js

pedrobaeza (approved):

pedrobaeza:
Release done

PR #301, opened 2025-04-21
[IMP] Add QUnit as global in eslint.config for >=18.0

CarlosRoca13:
In >= 18.0 QUnit is added as global https://github.com/odoo/odoo/blob/4b6770ec30bd00213674df2b075c77f33a0e65d9/addons/web/static/src/%40types/libs.d.ts#L5

Added the key on globals to avoid no-undef linter error.


pedrobaeza (approved):
Coming from https://github.com/OCA/web/pull/3145#discussion_r2052027523

chienandalu (approved):

PR #303, opened 2025-06-12
Fix 12.0 version container deprecation 

HviorForgeFlow:
- ubuntu-20.04 was deprecated few months ago. Use ubuntu-22.04 by default
- ~~python version needs to be fixed to 3.6.15 to be reconized by ubuntu-22.04~~ (Use py38 instead)


mostafabarmshory (approved):

rousseldenis:
@sbidoul

HviorForgeFlow:
For pre-commit I couldn't fine any solution with py36. the only one was using py38 to manage this part. Any idea on how keep py36?

PR #306, opened 2025-07-02
[IMP] bump maintainer tools to the latest version

ivs-cetmix:
Why? Because there are a lot of changes that are not reflected properly and may create a mess if different maintainer tools versions are used in OCA bot and repo template.
For example, in one of the commits we rename 'licence' into 'license'. But because pre-commit is pinned to the commit before this change it is renaming it back again and again making the commit history dirty. And although this was not yet spotted in the OCA repos, some other repos using OCA tools are affected.


pedrobaeza (approved):
I agree. We will need a massive repo update after this.

sbidoul (approved):

PR #310, opened 2025-09-15
Add Odoo 19

etobella:


jdoutreloux:
In `{% if ci == 'Travis' %}.travis.yml{% else %}.t2d.yml{% endif %}.jinja` the minimum python version is pinned. Shouldn't it be adapted as for now in the file it's 3.8 but min version for 19.0 is 3.10 ?

etobella:
@jdoutreloux travis is not being used anymore in OCA. It is kept for historical reasons, but it is not maintained. Actually, it has the same problem with Odoo 18

jdoutreloux:
@etobella I didn't know that, thanks for the explanation.

pedrobaeza (commented):
I think we should bump the respective library versions (ruff, pylint, etc) at the same time.

etobella:
libraries updates. Maybe we can merge it.

.github/workflows/test.yml
```diff
@@ -32,6 +32,9 @@ jobs:
           - odoo-version: 18.0
             python-version: "3.12"
             machine: ubuntu-24.04
+          - odoo-version: 19.0
+            python-version: "3.12"
```
pedrobaeza, line 36: Odoo is using this Python version for runbot?

PR #311, opened 2025-09-15
[19.0] upgrade  pre-commit hooks to latest version available

lmignon:
* [ ] includes #310 
* [x] waiting/blocked by https://github.com/OCA/pylint-odoo/pull/544

### DIFF
--- a/src/.pylintrc-mandatory.jinja
+++ b/src/.pylintrc-mandatory.jinja
@@ -106,6 +106,31 @@ enable=anomalous-backslash-in-string,
 {%- if odoo_version >= 16 %},
     external-request-timeout
 {%- endif %}
+{%- if odoo_version >= 19 %},
+    bad-builtin-groupby,
+    category-allowed,
+    deprecated-name-get,
+    deprecated-odoo-model-method,
+    inheritable-method-lambda,
+    inheritable-method-string,
+    invalid-email,
+    manifest-behind-migrations,
+    manifest-data-duplicated,
+    missing-odoo-file,
+    no-raise-unlink,
+    no-search-all,
+    no-wizard-in-models,
+    prohibited-method-override,
+    prefer-env-translation,
+    translation-format-interpolation,
+    translation-format-truncated,
+    translation-fstring-interpolation,
+    translation-not-lazy,
+    translation-too-few-args,
+    translation-too-many-args,
+    translation-unsupported-format,
+    no-write-in-compute
+{%- endif %}
 {%- block optional_checks %}{% endblock %}
 
 [REPORTS]
--- a/src/.pylintrc.jinja
+++ b/src/.pylintrc.jinja
@@ -26,6 +26,7 @@
     old-api7-method-defined,
     redefined-builtin,
     too-complex,
-    unnecessary-utf8-coding-comment
+    unnecessary-utf8-coding-comment,
+    manifest-external-assets
 {% endblock %}
 {%- endif %}


src/.pre-commit-config.yaml.jinja
```diff
@@ -91,6 +91,25 @@
...
+  {%- set repo_rev.flake8_bugbear = "21.9.2" %}
+  {%- set repo_rev.globals = "16.0.0" %}
+  {%- set repo_rev.isort = "6.0.1" %}
+  {%- set repo_rev.maintainer_tools = "f9b919b9868143135a9c9cb03021089cabba8223" %}
+  {%- set repo_rev.nodejs = "22.9.0" %}
+  {%- set repo_rev.odoo_pre_commit_hooks = "v0.1.6" %}
+  {%- set repo_rev.pre_commit_hooks = "v6.0.0" %}
+  {%- set repo_rev.prettier = "3.6.2" %}
+  {%- set repo_rev.prettier_xml = "3.4.2" %}
+  {%- set repo_rev.pylint_odoo = "v9.3.14" %}
```
moylop260, line 109: Thank you @lmignon 

FYI https://github.com/OCA/pylint-odoo/pull/544

```suggestion
  {%- set repo_rev.pylint_odoo = "v9.3.15" %}
```

moylop260:
We have added new checks that could be helpful for this new Odoo version where the self.env._ should be used, better inheritable, and translation validation and so on

If you agree, consider enabling them in the pylintrc files.

* bad-builtin-groupby (W8155): Used builtin function itertools.groupby. Prefer odoo.tools.groupby instead. More info about https://github.com/odoo/odoo/issues/105376

* category-allowed (C8114): Category "%s" not allowed in manifest file.

* deprecated-name-get (E8146): 'name_get' is deprecated. Use _compute_display_name instead. More info at https://github.com/odoo/odoo/pull/122085.

* deprecated-odoo-model-method (W8160): %s has been deprecated by Odoo. Please look for alternatives.

* inheritable-method-lambda (E8148): Use %s=lambda self: self.%s() to preserve inheritability. More info at https://github.com/OCA/odoo-pre-commit-hooks/issues/126

* inheritable-method-string (E8147): Use string method name "%s" to preserve inheritability. More info at https://github.com/OCA/odoo-pre-commit-hooks/issues/126

* invalid-email (R8181): Invalid email "%s"

* manifest-behind-migrations (E8145): Manifest version (%s) is lower than migration scripts (%s)

* manifest-data-duplicated (W8125): The file "%s" is duplicated in lines %s from manifest key "%s"

* manifest-external-assets (W8162): Asset %s should be distributed with module's source code. More info at https://httptoolkit.com/blog/public-cdn-risks/

* missing-odoo-file (C8115): Missing %s %sfile

* no-raise-unlink (E8140): No exceptions should be raised inside unlink() functions

* no-search-all (W8163): Using an empty domain %s([]) without a limit will load all records, may impact performance.

* no-wizard-in-models (C8113): No wizard class for model directory. See the complete structure https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#complete-structure

* no-write-in-compute (E8135): Compute method calling write. Use update instead.

* prefer-env-translation (W8161): Better using self.env._ More info at https://github.com/odoo/odoo/pull/174844

* prohibited-method-override (W8107): Prohibited override of "%s" method.

* translation-format-interpolation (W8302): Use %s formatting in odoo._ functions

* translation-format-truncated (E8301): Logging format string ends in middle of conversion specifier

* translation-fstring-interpolation (W8303): Use %s formatting in odoo._ functions

* translation-not-lazy (W8301): Use %s formatting in odoo._ functions

* translation-too-few-args (E8306): Not enough arguments for odoo._ format string

* translation-too-many-args (E8305): Too many arguments for odoo._ format string

* translation-unsupported-format (E8300): Unsupported odoo._ format character %r (%#02x) at index %d


See more [examples](https://github.com/OCA/pylint-odoo/blob/main/README.md#examples)

moylop260:
> What's the motivation behind 'No exceptions should be raised inside
unlink() functions' How can you prevent unlinking records if some
conditions are not met?

Odoo added a new decorator `api.ondelete`, more info here:
 - https://github.com/odoo/odoo/commit/1c8a95809852baefa851985507a07ac7f6465dc7

> About 'no-write-in-compute (E8135): Compute method calling write. Use
update instead.' Can you elaborate why?

@pedrobaeza request from https://github.com/OCA/pylint-odoo/issues/356

lmignon:
I added the new proposed rules expect 'no-write-in-compute (E8135): Compute method calling write. Use update instead.' The proposed solution too restrictive. We should also suggest to write direct assignment.

pedrobaeza:
@lmignon the linter was redacted that way because people usually uses `write` when they want to assign more than one value in one line. Something like:

```python
record.write({"field1": value1, "field2": value2})
```

So that should be converted into:

```python
record.update({"field1": value1, "field2": value2})
```

instead of:

```python
record.field1 = value1
record.field2 = value2
```

Residual cases like `record.write({"field": value}` to be converted to `record.field = value` is not covered, but as said, I think it's very residual, don't you think?

src/.pylintrc-mandatory.jinja
```diff
@@ -106,6 +106,30 @@ enable=anomalous-backslash-in-string,
 {%- if odoo_version >= 16 %},
     external-request-timeout
 {%- endif %}
+{%- if odoo_version >= 19 %},
+    bad-builtin-groupby,
+    category-allowed,
+    deprecated-name-get,
+    deprecated-odoo-model-method,
+    inheritable-method-lambda,
+    nheritable-method-string,
```
moylop260, line 115:
```suggestion
    inheritable-method-string,
```

lmignon: good catch: fixed

lmignon:
> @lmignon the linter was redacted that way because people usually uses `write` when they want to assign more than one value in one line. Something like:
> 
> ```python
> record.write({"field1": value1, "field2": value2})
> ```
> 
> So that should be converted into:
> 
> ```python
> record.update({"field1": value1, "field2": value2})
> ```
> 
> instead of:
> 
> ```python
> record.field1 = value1
> record.field2 = value2
> ```
> 
> Residual cases like `record.write({"field": value}` to be converted to `record.field = value` is not covered, but as said, I think it's very residual, don't you think?

In my opinion, it's about spreading best practices. You're right, “write” should be avoided/prohibited, but I think it's also important to promote the most appropriate solution. Let's see what others are thinking...

PR #314, opened 2025-09-19
[IMP] upload screenshots as artifacts if they exist and tests fail

hbrunn:
given this only runs on test failures and doesn't do anything if there are no screenshots, I think it can be unconditional.

You can check out how this looks [here](https://github.com/OCA/crowdfunding/actions/runs/17851749903/job/50761657310?pr=6#step:9:1) and [here](https://github.com/OCA/crowdfunding/actions/runs/17851749903?pr=6) (scroll down to artifacts), was already helpful in that case.


hbrunn:
this one would be a tragedy if we lost it through inaction

PR #324, opened 2025-10-27
Push pot without pat

sbidoul:
Reopening #290 as I still want to do this at some point.


sbidoul:
I still want to do this at some point. Not stale.

PR #326, opened 2025-10-28
[FIX] Set python minor version on pre-commit

david-banon-tecnativa:
I was having issues with flake on 16.0, because it was using pythyon 3.13 and it seems to be incompatible with the current config. This sets the python version to the same used to run CI to avoid this problem.

@Tecnativa

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -1,3 +1,4 @@
+{% from 'src/.pre-commit-config.yaml.jinja' import repo_rev with context %}
 name: pre-commit
 
 on:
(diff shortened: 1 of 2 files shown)


yajo (approved):

hbrunn (approved):

rvalyi (approved):

david-banon-tecnativa:
I'm not really sure, but i think it will help. While on the original issue i couldn't even run pre-commit succesfully. 2 different python versions might be able to run it successfully but produce different results. So forcing the same python version might fix some of the pollution.

sbidoul:
I also think we should do this.

Could you update `src/.github/workflows/pre-commit.yml.jinja` to use the same variable?

david-banon-tecnativa:
Done @sbidoul

pedrobaeza:
ping @sbidoul

carlosdauden (approved):

pedrobaeza:
@etobella It seems Stéphane is on holidays, so maybe you can merge and release the version?

etobella (approved):

PR #330, opened 2026-02-09
[IMP] pre-commit: Add refurb

SirPyTech:
Refurb (https://github.com/dosisod/refurb) is:
> A tool for refurbishing and modernizing Python codebases.

This PR proposes to add it as a `pre-commit`  check for repositories, default for future versions.

The `migrations` folder is excluded because it raises an error like:
```
l10n_it_account_stamp/migrations/18.0.1.0.0/post-migration.py: error: Duplicate module named "post-migration" (also at "l10n_it_account/migrations/18.0.1.0.0/post-migration.py")
l10n_it_account_stamp/migrations/18.0.1.0.0/post-migration.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
l10n_it_account_stamp/migrations/18.0.1.0.0/post-migration.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
```
Excluding them is a small price to pay, to have the whole rest of the codebase improved in my opinion.

Let me know what you think!

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -141,6 +141,14 @@ additional_ruff_rules:
   help: List of additional ruff rules to enable.
   when: "{{ use_ruff }}"
 
+use_refurb:
+  default: "{% if odoo_version < 20.0 -%}no{% else %}yes{% endif %}"
+  type: bool
+  help:
+    Use refurb (https://github.com/dosisod/refurb) for refurbishing and modernizing
+    Python codebase.
+  when: "{{ odoo_version >= 14.0 }}"
+
 rebel_module_groups:
   type: yaml
   default: []
--- a/copier.yml
+++ b/copier.yml
@@ -100,6 +100,14 @@ additional_ruff_rules:
   help: List of additional ruff rules to enable.
   when: "{{ use_ruff }}"
 
+use_refurb:
+  default: "{% if odoo_version < 20.0 -%}no{% else %}yes{% endif %}"
+  type: bool
+  help:
+    Use refurb (https://github.com/dosisod/refurb) for refurbishing and modernizing
+    Python codebase.
+  when: "{{ odoo_version >= 14.0 }}"
+
 rebel_module_groups:
   type: yaml
   default: []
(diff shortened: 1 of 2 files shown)


HekkiMelody (approved):
LGTM!

While I think this is the best approach, and I really like the tool, I should mention that refurb can feel a bit slow on large codebases. A potential, less-disruptive alternative is to enable the rules inspired by refurb that have been imported to ruff https://docs.astral.sh/ruff/rules/#refurb-furb

However, the rules available in ruff are only a subset of those available in refurb https://github.com/dosisod/refurb/blob/master/docs/checks.md which is why I think this is best.

SirPyTech:
Rebased and updated `refurb`s revision.

quirino95 (approved):
Code and functional review: LGTM!

HekkiMelody (approved):

sbidoul:
Since ruff [seems to implement many refurb rules](https://github.com/astral-sh/ruff/issues/1348) already, could we enable (some of) those instead to avoid adding one more tool?

SirPyTech:
> Since ruff [seems to implement many refurb rules](https://github.com/astral-sh/ruff/issues/1348) already, could we enable (some of) those instead to avoid adding one more tool?

Thanks a lot @sbidoul for having a look, I already answered a similar observation in https://github.com/Tecnativa/doodba-copier-template/pull/599#issuecomment-3985004067, copying here the relevant part:
> `ruff`'s list of checks is only a subset of all the `refurb` checks: there are only 36 in https://docs.astral.sh/ruff/rules/#refurb-furb in contrast with the 93 in https://github.com/dosisod/refurb/blob/master/docs/checks.md.
That's probably a matter of taste: have few checks autofixed or have all the checks suggested? I personally would go with the latter.

Please have a look at the linked discussion for more clarifications.

PR #336, opened 2026-02-23
Fix pre-commit job pip cache config

sbidoul:


pedrobaeza (approved):

sbidoul:
Please don't merge before testing on an affected repo. I'm not sure this is correct.

src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -29,6 +29,7 @@ jobs:
           python-version: "3.11"
 {%- endif %}
           cache: 'pip'
+          cache-dependency-path: '.pre-commit-config.yaml.jinja'
```
huguesdk: should the `.jinja` suffix be there?

sbidoul: Nope, thanks!

SirPyTech:
> Here a requirements.txt would not make sense as a cache key. Since we are in pre-commit, `.pre-commit-config.yaml` could be a good cache key.

Right: the cache is used for the libraries installed for `pre-commit`'s hooks so it makes more sense to use its configuration as a cache key.

I'm not really convinced that using only the hook's versions is a good cache key because that would misbehave when a dependency of the hook is changed: in a new (local) environment the new dependency would be used, but the github action would use the cached versions?
Or maybe I should just look better into how the cache works for the github action.

It makes sense to use a `requirements` file as a cache key because it should have all the libraries with their versions pinned but there is no such thing for `pre-commit` that I know of.
Maybe we shouldn't use cache at all for `pre-commit`?

PR #339, opened 2026-04-04
Remove Travis support

sbidoul:
OCA does not use Travis anymore so let's shed some weight.


src/README.md.jinja
```diff
@@ -12,9 +12,6 @@
 {%- else %}
 <!-- /!\ Non OCA Context : Set here the badge of your runbot / runboat instance. -->
 {%- endif %}
-{%- if ci == "Travis" %}
-[![Build Status](https://travis-ci.com/{{ org_slug }}/{{ repo_slug }}.svg?branch={{ odoo_version }})](https://travis-ci.com/{{ org_slug }}/{{ repo_slug }})
-{%- else %}
```
legalsylvain, line 17: I guess you should also remove the endif line 20.!

legalsylvain (commented):
a typo inline. Otherwise, LGTM.

legalsylvain (approved):
Thanks !

/ocabot merge nobump

mostafabarmshory:
Dear @sbidoul

I use OCA maintainer tools to manage my projects, so it would be good to take this into account in future developments.

However, thanks to your work, this has been improved.

rvalyi:
Merging as it is approved. Also I think it makes more sense to use the rebase+merge button here, right?

PR #340, opened 2026-04-04
Use standard-based declaration of test dependencies

sbidoul:
Now that dependency groups are [standardized](https://packaging.python.org/en/latest/specifications/dependency-groups/) we can leverage it to have a tool-agnostic configuration of test dependencies.


legalsylvain (approved):
Lgtm. Do you want some tests, or did you tested it ?

sbidoul:
> Lgtm. Do you want some tests, or did you tested it ?

I have tested locally, and CI is green.

sbidoul:
@yajo is this one ok for you?

.gitignore
```diff
@@ -1,3 +1,6 @@
+# ignore uv.lock as we want to test with the latest versions
+uv.lock
```
yajo: why ignore it?

sbidoul: Because I think it's better to run tests with the latest copier version and discover problems early. Indeed users of the template will not use the lock file so I feel it's better to test without locked dependencies to be closer to what users will have.

yajo: IMHO it's better to have lock files to avoid random failures from unknown sources. We can have a job that auto-updates that lock file under a certain schedule to narrow those updates.

Template users will use the lock file if provided AFAIK.

sbidoul: Alright I've added uv.lock and a dependabot config.

> Template users will use the lock file if provided AFAIK.

What do you mean with that?

yajo: > Alright I've added uv.lock and a dependabot config.
> 
> > Template users will use the lock file if provided AFAIK.
> 
> What do you mean with that?

You said:



> users of the template will not use the lock file

However, if they check out this repo, they will use the lock file if present.

yajo (approved):

.gitignore
sbidoul: Ah yes, but those would be developers of the template, not regular users.

sbidoul:
Thanks for the review.

PR #351, opened 2026-05-07
[FIX] Rename test.yaml to render correctly

carlos-lopez-tecnativa:
With the removal of Travis support, the CI question no longer exists, so this file is no longer rendered. In new repositories, test.yaml is not copied correctly.

Complementary to https://github.com/OCA/oca-addons-repo-template/pull/339

See https://github.com/OCA/route-planning/tree/18.0/.github/workflows 

@sbidoul @rvalyi @pedrobaeza could you please review this?


pedrobaeza (approved):

sbidoul:
Thanks!

PR #355, opened 2026-05-22
Add postgres image parameter

etobella:
This is useful for example for field_vector


etobella:
@lmignon as original author of field_vector, I assume this might be interesting to you

lmignon (approved):
Good Idea @etobella Thank you!

zamberjo (approved):
https://github.com/OCA/oca-addons-repo-template/pull/354 attempts to address this issue. I prefer this solution.

etobella:
@sbidoul can we merge and add a new copier version?

PR #358, opened 2026-06-17
Cancel in-progress CI runs on new pushes to the same branch

### LINKED ISSUE #125: GitHub actions: cancel previous builds of same branch
When there are successive pushes to a given branch or PR, cancel running jobs on the same branch.

dnplkndll:
Adds a `concurrency` group with `cancel-in-progress: true` to the generated `test.yml` and `pre-commit.yml` so successive pushes to the same branch/PR cancel the previous still-running jobs

Closes #125

### DIFF
--- a/src/.github/workflows/pre-commit.yml.jinja
+++ b/src/.github/workflows/pre-commit.yml.jinja
@@ -10,6 +10,10 @@ on:
       - "{{ odoo_version }}"
       - "{{ odoo_version }}-ocabot-*"
 
+concurrency:
+  group: {% raw %}${{ github.workflow }}-${{ github.ref }}{% endraw %}
+  cancel-in-progress: true
+
 jobs:
   pre-commit:
     runs-on: ubuntu-22.04
--- a/src/.github/workflows/test.yml.jinja
+++ b/src/.github/workflows/test.yml.jinja
@@ -9,6 +9,10 @@ on:
       - "{{ odoo_version }}"
       - "{{ odoo_version }}-ocabot-*"
 
+concurrency:
+  group: {% raw %}${{ github.workflow }}-${{ github.ref }}{% endraw %}
+  cancel-in-progress: true
+
 {%
 set IMAGES = {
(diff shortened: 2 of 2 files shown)


zamberjo (approved):

src/.github/workflows/test.yml.jinja
```diff
@@ -9,6 +9,10 @@ on:
       - "{{ odoo_version }}"
       - "{{ odoo_version }}-ocabot-*"
 
+{% raw %}concurrency:
```
legalsylvain: Could you explain the syntax ? I dont get what is raw ? 

thanks !

ref https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-concurrency-and-the-default-behavior

dnplkndll: the {% raw %} prevents jinja eval like `<![CDATA[ … ]]>`

src/.github/workflows/pre-commit.yml.jinja
```diff
@@ -10,6 +10,10 @@ on:
       - "{{ odoo_version }}"
       - "{{ odoo_version }}-ocabot-*"
 
+{% raw %}concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true{% endraw %}
+
```
legalsylvain:
```suggestion
concurrency:
  group: {% raw %}${{ github.workflow }}-${{ github.ref }}{% endraw %}
  cancel-in-progress: true
```

nitpicking.

src/.github/workflows/test.yml.jinja
legalsylvain: oh thanks !

legalsylvain (approved):
thanks ! 

This should reduce (very slightly) resource consumption.

david-banon-tecnativa (approved):

PR #359, opened 2026-06-20
[REF] pylintrc-mandatory: enable deprecated module checks

moylop260:
Enable pylint's deprecated-module check in the mandatory rcfile used by the blocking pre-commit hook for Odoo 13.0 and newer.

This makes generated OCA repositories fail fast when forbidden imports such as pdb, pudb, ipdb, or bs4 are introduced, instead of leaving the check only in the optional IDE-oriented pylint configuration.

For Odoo 20.0 and newer, also mark pytz as deprecated so repositories generated from this template follow the upstream direction taken in Odoo saas-19.1.

That branch removed pytz usage in Odoo commit:
 - https://github.com/odoo/odoo/commit/6fdb36716aac5d2b084fb0faa94ccb2f3ee7f99c

Keep the scope limited to 13.0+ templates, where src/.pylintrc-mandatory.jinja is the source used to render both .pylintrc-mandatory and the inherited optional .pylintrc. Legacy <=12.0 templates are intentionally left unchanged.

Extend the bootstrap test to assert the rendered [IMPORTS] section and the version-specific pytz addition.

### DIFF
--- a/src/.pylintrc-mandatory.jinja
+++ b/src/.pylintrc-mandatory.jinja
@@ -5,6 +5,9 @@
 load-plugins=pylint_odoo
 score=n
 
+[IMPORTS]
+deprecated-modules=pdb,pudb,ipdb,bs4{%- if odoo_version >= 20 %},pytz{%- endif %}
+
 [ODOOLINT]
 {%- if odoo_version < 16 %}
 readme_template_url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
@@ -46,6 +49,7 @@ enable=anomalous-backslash-in-string,
     assignment-from-none,
     attribute-deprecated,
     dangerous-default-value,
+    deprecated-module,
     development-status-allowed,
     duplicate-key,
     eval-used,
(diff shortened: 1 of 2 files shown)


moylop260:
@sbidoul @hbrunn @StefanRijnhart 

Could you review it, please?


Odoo v19.1 have deleted pytz dependency

A faster fail is enabling this module as deprecated

hbrunn (approved):

zamberjo (approved):

StefanRijnhart (approved):
Thanks!

moylop260:
@sbidoul 

Approved

Could you merge it, please?

sbidoul:
Hi, I'm back from holidays.

For these kind of generic python checks I have a preference to use ruff and keep pylint for Odoo specific checks.

Ruff rules [banned-api](https://docs.astral.sh/ruff/rules/banned-api/) and [debugger](https://docs.astral.sh/ruff/rules/debugger/) do this.

A question: what's wrong with `bs4` ?

moylop260:
Hi @sbidoul 

Ruff rules sound better! 👍

> A question: what's wrong with `bs4` ?

Odoo forces to use `lxml` so you don't need use `bs4` for Odoo projects in almost of the cases since that the lxml cover the methods

But a few developers install this extra requirement to do the same than lxml can do it

PR #360, opened 2026-06-24
[FIX] prettier 3.x: actually use plugin-xml

hbrunn:
while working on formatting some xml files, I noticed that prettier doesn't actually load plugin-xml in versions 3.x. You can check this by adding `--xml-whitespace-sensitivity=strict` to args, this will fail with the current configuration.

This seems to be https://github.com/prettier/prettier/issues/15085

Note that this will cause large scale reformatting on existing files, because the html formatter that's used otherwise formats xml slightly differently (which is why we want plugin-xml in the first place though)

@yajo I've seen you've worked on this before, what is your conclusion here?

### DIFF
--- a/src/.pre-commit-config.yaml.jinja
+++ b/src/.pre-commit-config.yaml.jinja
@@ -227,8 +227,20 @@ repos:
     hooks:
       - id: prettier
         name: prettier (with plugin-xml)
+        {%- if odoo_version >= 18 %}
+        # hack: prettier version 3.x doesn't auto search for plugins, we need to pass
+        # them explicitly
+        entry:
+          bash -c "prettier --plugin=$NODE_PATH/@prettier/plugin-xml/src/plugin.js $0
+          $*"
+        {%- else %}
         entry: prettier
+        {%- endif %}
         args:
+        {%- if odoo_version < 18 %}
+          - --plugin=@prettier/plugin-xml
+        {%- endif %}
+          - --xml-whitespace-sensitivity=strict
           - --write
           - --list-different
           - --ignore-unknown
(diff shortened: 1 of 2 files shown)


yajo:
This is cleaner, did you try it? https://github.com/prettier/plugin-xml/issues/17#issuecomment-1767792202

hbrunn:
this fails with the error message linked above (Cannot find package plugin-name imported from ...), that's why I think we need to pass the absolute path to the plugin file

yajo:
Yes, it's done there. Just with `--plugin`

hbrunn:
does this work locally for you? here it doesn't

yajo:
Yes, since long ago. [Check it on doodba](https://github.com/Tecnativa/doodba-copier-template/blob/842976dc45d4dd96c1938b89ea5636790d7d8a45/template/.pre-commit-config.yaml.jinja#L115-L125).

hbrunn:
yeah you're using prettier 2.x. this is 3.x

hbrunn:
I've added a test to check if xml formatting actually happens and has the expected results

yajo (approved):

sbidoul:
@hbrunn @yajo I just checked on mis-builder 18.0 which has the latest template, it seems to work correctly? I added a space before an XML attribute and pre-commit correctly reformatted it.

It has a `prettier.config.cjs` which loads `plugin-xml` and has `xmlWhitespaceSensitivity: "preserve"` which is the one we want.

sbidoul:
See https://github.com/OCA/mis-builder/pull/811

hbrunn:
thank you @sbidoul for checking, I missed the config file. When you pass this as parameter, it fails, but setting it in the config file works indeed. So good enough as-is, closing

PR #361, opened 2026-07-02
[REF] pre-commit: Enable 'translation-injection' pylint-odoo check for odoo v19.0+

moylop260:
If a repo raises this check so it is a security risk related to:
 - https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

More info:
 - https://github.com/OCA/pylint-odoo/pull/572

Thanks to @SirPyTech contributions:
  - https://github.com/OCA/pylint-odoo/issues/571
  - https://github.com/OCA/odoo-pre-commit-hooks/pull/192


yajo (approved):

SirPyTech (approved):
Thanks!
As you spotted in https://github.com/OCA/pylint-odoo/pull/575 this might not be a _big_ security risk, but it's good in my opinion to have this check.

Why not enable it in older versions too?

yajo:
> Why not enable it in older versions too?

There is a little friction in the community when enabling new checks on old versions. Some think that it gives more problems than help.

Let's merge this and you can propose the older versions in a future PR if you want.

Thanks all!

PR #362, opened 2026-07-13
[IMP] checklog-odoo: ignore warning about website.list_hybrid in search

hbrunn:
Searching on the website emits a [warning](https://github.com/OCA/crowdfunding/actions/runs/29084229112/job/86334020547?pr=25#step:8:395) because of the t-key attribute in https://github.com/OCA/OCB/blob/18.0/addons/website/views/website_templates.xml#L2738.

Won't be fixed in upstream as of https://github.com/odoo/odoo/pull/179017#issuecomment-4934713678, so we need to ignore this to be able to test anything involving the search frontend


pedrobaeza (approved):
Well, I have written in the Odoo PR to support its merge, but if not, we need this alternative.

pedrobaeza:
It's being merged in odoo/odoo, so this is not finally needed.

hbrunn:
will they actually do this for 18, 19 too?

pedrobaeza:
Yes, they hadn't blacklist robodoo for their automatic fw-port mechanism, so it will take a bit, but we expect it to have it.

PR #363, opened 2026-07-18
[IMP] pre-commit: Update pylint-odoo to v10.0.8 to enable deprecated-self-cr only for Odoo v19.0+

moylop260:
Bump the pylint-odoo hook to v10.0.8
(see https://github.com/OCA/pylint-odoo/pull/577)
and add the new `deprecated-self-cr` check
to the mandatory checks list in `.pylintrc-mandatory` to flag deprecated usage of `self._cr`.


yajo (approved):

sbidoul:
Will this make many repos red when we deploy it? As always, adding new check to existing branches is a burden for contributors.

moylop260:
Hi @sbidoul 

Good question

I thought that the CI should be yellow since that it is a warning from Odoo's logger:
 - https://github.com/odoo/odoo/blob/76573511fd58dac5bc0bcb0/odoo/orm/models.py#L5912

But I just downloaded all the OCA's repositories 19.0 (since that this lint only affects this version 19.0+)

And the following projects could be affected with
 - `W8165: Use "self.env.cr" instead of "self._cr" (deprecated since 19.0) (deprecated-self-cr)`

## report-print-send

### base_report_to_printer [installable: True](https://github.com/OCA/report-print-send/blob/55a891db2342ecd0d0b0ddc208ebe03a40da4558/base_report_to_printer/__manifest__.py#L34)
 - https://github.com/OCA/report-print-send/blob/55a891db2342ecd0d0b0d/base_report_to_printer/models/ir_actions_report.py#L151C23

## storage

### fs_storage_ms_drive [installable: False](https://github.com/OCA/storage/blob/af1a3beeb324082e457bb168d20d13704f5df323/fs_storage_ms_drive/__manifest__.py#L13)
  - https://github.com/OCA/storage/tree/af1a3beeb324082e4/fs_storage_ms_drive/models/fs_storage.py#L33C33

### storage_file [installable: False](https://github.com/OCA/storage/blob/af1a3beeb324082e457bb168d20d13704f5df323/storage_file/__manifest__.py#L15)
 - https://github.com/OCA/storage/tree/af1a3beeb324082e4/storage_file/models/storage_file.py#L204C9 
 - https://github.com/OCA/storage/tree/af1a3beeb324082e4/storage_file/models/storage_file.py#L209C30


Notice the `storage` project has the both modules `installable: False`

Not big impact here since that they are not already migrated


`report-print-send` is `installed: True` and it is affected but only if someone decides to run `copier update` in the project

I have checked the last update, and it was never updated before, it only have the initial commit:
 - https://github.com/OCA/report-print-send/commits/55a891db2342ecd0d0b0ddc208ebe03a40da4558/.copier-answers.yml

But if someone wants to update the copier, maybe they want to have the latest changes including the new lints to improve the code

> Will this make many repos red when we deploy it? 

No, will not.

0 repositories will be made red when we merge this PR.

Also, it is important to comment the following real warning in the environment using the module `base_report_to_printer` 🟨 `WARNING db py.warnings: report-print-send_19.0/base_report_to_printer/models/ir_actions_report.py:151: DeprecationWarning: Deprecated since 19.0, use self.env.cr directly`

 - Screenshot 2026-07-20 at 4 43 05 p m

So, even I think it should be fixed

and the fix is even easier

```diff
diff --git a/base_report_to_printer/models/ir_actions_report.py b/base_report_to_printer/models/ir_actions_report.py
index f469840..b0ce86f 100644
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -148,7 +148,7 @@ class IrActionsReport(models.Model):
                 return

     def print_document_threaded(self, report_id, record_ids, data):
-        with Registry(self._cr.dbname).cursor() as cr:
+        with Registry(self.env.cr.dbname).cursor() as cr:
             self = self.with_env(self.env(cr=cr))
             report = self.env["ir.actions.report"].browse(report_id)
             report.print_document(record_ids, data)
```


Note: The CI is not reporting this warning because the unittest is not covering this line
 - 
Screenshot 2026-07-20 at 4 57 47 p m

- [coverage report](https://app.codecov.io/github/OCA/report-print-send/commit/55a891db2342ecd0d0b0ddc208ebe03a40da4558/blob/base_report_to_printer/models/ir_actions_report.py?dropdown=coverage)

WDYT?

sbidoul (approved):
Thanks for the analysis @moylop260 !

sbidoul:
> I thought that the CI should be yellow since that it is a warning from Odoo's logger:

Yeah the python deprecation warnings are probably not surfaced. A while ago I had tried that but it was too noisy because Odoo was using too many deprecated python features. I don't know what's the situation today. Probably worth investigating again.

Issue #335, opened 2026-02-22
Using renovate to run copier updates


sbidoul:
[Renovate](https://docs.renovatebot.com) has the capability to [run copier update ](https://docs.renovatebot.com/modules/manager/copier/).

I'm currently experimenting with a per [repo config like this](https://github.com/OCA/mis-builder/blob/18.0/renovate.json), which needs to be defined on the main branch of each repo.

We'd need to self-host renovate, either on the OCA server, or using a org-wide scheduled GitHub action, because our copier template needs `--trust`. Need to think about the security implications.

WIth a renovate config like this

```json
{
  "plaform": "github",
  "allowScripts": true,
  "autodiscover": false,
  "repositories": [
    "OCA/mis-builder"
  ],
  "username": "oca-renovate-bot",
  "gitAuthor": "oca-renovate-bot <oca-renovate-bot@odoo-community.org>"
}
```

it produced pull requests such as https://github.com/OCA/mis-builder/pull/763, https://github.com/OCA/mis-builder/pull/767, https://github.com/OCA/mis-builder/pull/768.

To be continued....

yajo:
> because our copier template needs `--trust`

I think none of the tasks actually make sense these days and could be removed, so `--trust` can be removed too.

sbidoul:
The unsafe tasks are removal of obsolete files:

```
_tasks:
  - rm -f {% if ci != "Travis" %}.travis.yml{% endif %} .t2d.yml
  - "{% if odoo_version >= 18 %}rm -f .prettierrc.yml .eslintrc.yml{% endif %}"

_migrations:
  - version: v1.3.6
    after:
      - rm -f CONTRIBUTING.md
```

Some are important.

We could add a pre-commit [fail](https://pre-commit.com/#fail) entry on such files so maintainers have to remove them manually.

Or, @yajo, do you know if there is something in copier to delete files without `--trust`?

yajo:
Since Copier 6, removal of files that disappeared in the template is automatic. Relevant commit: https://github.com/copier-org/copier/commit/23dd5dbf8f727ad50ac8c388f2312969ca5cee83

Also, `--trust` is only needed in migrations that pass through that specific template update.

The template requires Copier 9: https://github.com/OCA/oca-addons-repo-template/blob/38b45f651dd8b51e1194722023c449735e61a8d9/copier.yml#L4

So, in short, you can just remove all that code today 😉

sbidoul:
[...]

PR #364, opened 2026-07-19
Add renovate.json to auto update the template

sbidoul:
Renovate [reads the renovate.json configuration from the default branch](https://docs.renovatebot.com/configuration-options/#basebranchpatterns), and applies it to the listed branches. Here we apply it to the current, next, and previous 3 Odoo series.

towards #335


sbidoul:
I'm merging this and will apply it to a few repos to see how it goes.

PR #369, opened 2026-09-17
20.0 support

feketemihai:


etobella:
https://github.com/OCA/oca-addons-repo-template/blob/6e74a6adbafe4d6f8d8b59291313bc446cd26e91/src/.github/workflows/test.yml.jinja#L133

On 20 we require postgres 16 according to odoo documentation

sbidoul:
To be continued in #370

